PR #31, opened 2015-04-24
Don't clear all cached offers everytime MesosNimbus receives a new set o...

erikdw:
...f offers

The previous behavior prevents accumulating enough offers to launch topologies,
since Mesos doesn't offer *all* resources every time resourceOffers() is called.

We still decline some of the cached offers every 120 seconds by default,
when RotatingMap.rotate() is invoked by a timer, so we don't permanently
hoard offers.


tnachen:
LGTM

PR #32, opened 2015-04-24
Improve logging in MesosNimbus

erikdw:
The original logging done in MesosNimbus has a high signal-to-noise
ratio due to using standard Protobuf toString methods.

This change is largely about pulling out important information from the
Protobuf objects and then printing the info in a more condensed format,
using JSON for any objects that have multiple internal fields.
This allows us to print more information in many less lines than before.

The goal is to allow introspecting the behavior of the MesosNimbus,
especially with respect to its handling of Offers and interactions with
the core nimbus code thru the INimbus interface methods.

Also did some minor clean up of various log lines in MesosNimbus.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -111,9 +111,9 @@
         <version>1.1</version>
       </dependency>
       <dependency>
-        <groupId>commons-lang</groupId>
-        <artifactId>commons-lang</artifactId>
-        <version>2.6</version>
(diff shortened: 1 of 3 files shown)


erikdw:
@tnachen : please take a gander when you have a chance.

tnachen:
LGTM

PR #33, opened 2015-04-24
Use CountDownLatch instead of Semaphore for awaiting completion of drive...

erikdw:
...r registration

This creates a cleaner interaction between the inner classes which
implement the Mesos interfaces (Scheduler, Executor) by avoiding
passing semaphore objects in the constructors.


erikdw:
@tnachen Please :eyes: this, thx!

PR #36, opened 2015-04-28
use docker to run nimbus (framework) and supervisor

tysonnorris:
This is an initial pass at streamlining the building and running of the framework by way of docker. 
Other minor changes include:
- update storm version to 0.9.4
- use slf4j


erikdw:
In general, can you please merge nicely  with @tnachen 's changes in this pull request: https://github.com/mesos/storm/pull/34

i.e., he also messed with the building logic, and it seems to conflict with what you've done here, so can you please rebase on top of his change, and then adjust accordingly?

src/storm/mesos/MesosNimbus.java
```diff
@@ -496,49 +508,76 @@ public void assignSlots(Topologies topologies, Map<String, Collection<WorkerSlot
...
+                          .setName("mem")
+                          .setType(Type.SCALAR)
+                          .setScalar(Scalar.newBuilder().setValue(executorMem))
+                          .setRole(memRole)))
                   .addResources(Resource.newBuilder()
-                      .setName("cpus")
-                      .setType(Type.SCALAR)
-                      .setScalar(Scalar.newBuilder().setValue(workerCpu))
-                      .setRole(cpuRole))
+                                    .setName("cpus")
```
tysonnorris: ugh sorry didn't realize the formatting was clobbered during this merge. will fix it

Dockerfile
```diff
@@ -1,22 +1,53 @@
-#
-# Dockerfile for Storm Mesos framework
-#
-FROM mesosphere/mesos:0.22.0-1.0.ubuntu1404
-MAINTAINER Timothy Chen <tnachen@gmail.com>
+#this is oracle java8 atop phusion baseimage
+FROM opentable/baseimage-java8:latest
```
tysonnorris: I agree the mesosphere image may be better to use here, will fix this

tysonnorris:
I found some more problems with cluster.xml that need to be delt with - will submit a new pull request asap.

erikdw:
@tysonnorris : cool -- I honestly am pretty ignorant about docker, so I'm not sure I actually grok what's being done here, but I wanna catch up.  Regarding formatting:  <philosophy>I don't use autoformatters, I don't believe in them</philosophy>.

Dockerfile
```diff
@@ -1,22 +1,53 @@
...
+RUN cat /etc/apt/sources.list.d/mesosphere.list
+
+#to make mesos repo available and install mesos:
+RUN apt-get update && apt-get install -y \
+    mesos \
+    python
+
+#install storm
+#inspired by https://github.com/wurstmeister/storm-docker/blob/master/storm/Dockerfile
+RUN wget -q -O - http://mirrors.sonic.net/apache/storm/apache-storm-0.9.4/apache-storm-0.9.4.tar.gz | tar -xzf - -C /opt
```
tnachen: If you remove the build release stuff and move it here, then users will have to use docker then right?
I think we should make this optional

tysonnorris: Yes correct. I was hoping to make the build process simpler, like only require "mvn clean package" (and for docker image, "docker build -t mesos-storm ."), but currently storm doesn't have the dist artifacts in mvn, just the libs etc.  
I'll leave my docker setup in my fork and if I think of some better way to allow non-docker and docker builds to coexist, will submit a new pull request.

Issue #35, opened 2015-04-28
Storm Authentication?


pdread100:
Is Storm authentication being worked? Plans? Ideas?

pdread100:
Some one off line answered some of my question, that is mesos storm should use the mesos authentication framework. So I will add this if no one else has or is doing so.

tnachen:
Please go ahead and help add framework authentication (principal, secret, etc) to this framework.

erikdw:
This has already been fixed by #37

PR #37, opened 2015-04-30
Mesos authentication

pdread100:
Coded and tested Mesos authentication to resolve issue [#35] (https://github.com/mesos/storm/issues/35)

### DIFF
--- a/src/storm/mesos/MesosNimbus.java
+++ b/src/storm/mesos/MesosNimbus.java
@@ -20,10 +20,12 @@ package storm.mesos;
 import backtype.storm.Config;
 import backtype.storm.scheduler.*;
 import backtype.storm.utils.LocalState;
+
 import com.google.common.base.Function;
 import com.google.common.base.Optional;
 import com.google.protobuf.ByteString;
 import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
 import org.apache.mesos.MesosSchedulerDriver;
 import org.apache.mesos.Protos.*;
@@ -36,6 +38,8 @@ import org.apache.mesos.Scheduler;
 import org.apache.mesos.SchedulerDriver;
 import org.json.simple.JSONValue;
 
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
@@ -58,6 +62,9 @@ public class MesosNimbus implements INimbus {
   public static final String CONF_MESOS_ALLOWED_HOSTS = "mesos.allowed.hosts";
   public static final String CONF_MESOS_DISALLOWED_HOSTS = "mesos.disallowed.hosts";
   public static final String CONF_MESOS_ROLE = "mesos.framework.role";
+  public static final String CONF_MESOS_PRINCIPAL = "mesos.framework.principal";
+  public static final String CONF_MESOS_SECRET_FILE = "mesos.framework.secret.file";
+
   public static final String CONF_MESOS_CHECKPOINT = "mesos.framework.checkpoint";
   public static final String CONF_MESOS_OFFER_LRU_CACHE_SIZE = "mesos.offer.lru.cache.size";
   public static final String CONF_MESOS_LOCAL_FILE_SERVER_PORT = "mesos.local.file.server.port";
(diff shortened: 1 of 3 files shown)


src/storm/mesos/MesosNimbus.java
```diff
@@ -97,57 +105,93 @@ public String getHostName(Map<String, SupervisorDetails> map, String nodeId) {
...
+      
     } catch (Exception e) {
       throw new RuntimeException(e);
     }
   }
-
+  
+  @SuppressWarnings("unchecked")
+  protected void initialize(Map conf, String localDir) throws Exception {
+
```
tnachen: Please kill this new line

src/storm/mesos/MesosNimbus.java
```diff
@@ -97,57 +105,93 @@ public String getHostName(Map<String, SupervisorDetails> map, String nodeId) {
...
   }
-
+  
+  @SuppressWarnings("unchecked")
+  protected void initialize(Map conf, String localDir) throws Exception {
+      
+      setConfig(conf);
+      setLocalState(new LocalState(localDir));
+
+      _allowedHosts = listIntoSet((List<String>)conf.get(CONF_MESOS_ALLOWED_HOSTS));
```
tnachen: I wonder if listIntoSet can be replaced with google's gauve Sets.newHashSet?

src/storm/mesos/MesosNimbus.java
```diff
@@ -97,57 +105,93 @@ public String getHostName(Map<String, SupervisorDetails> map, String nodeId) {
...
+  }
+  protected LocalState getLocalState() {
+      return _state;
+  }
+  protected void createLocalServerPort() {
+    Integer port = (Integer) _conf.get(CONF_MESOS_LOCAL_FILE_SERVER_PORT);
+    LOG.info("Using local port: " + port);
+    setLocalServerPort(Optional.fromNullable(port)); 
+  }
+  protected void setLocalServerPort(Optional<Integer> port) {
```
tnachen: Any reason why you added these getters and setters for a local variable?

src/storm/mesos/MesosNimbus.java
```diff
@@ -707,7 +751,58 @@ public void executorLost(SchedulerDriver driver, ExecutorID executor, SlaveID sl
...
+            .setCheckpoint(checkpoint);
+
+        if (id != null) {
+          finfo.setId(FrameworkID.newBuilder().setValue(id).build());
+        } 
+        
+        return finfo;
+    }
+    
+    private Credential checkForAuthentication(FrameworkInfo.Builder finfo) {
```
tnachen: the name checkForAuthentication doesn't seems to suggest it's returning Crendential. How about calling it getAuthentication that returns a Optional<Credential>?

src/storm/mesos/MesosNimbus.java
```diff
@@ -97,57 +105,93 @@ public String getHostName(Map<String, SupervisorDetails> map, String nodeId) {
...
+  }
+  protected void createLocalServerPort() {
+    Integer port = (Integer) _conf.get(CONF_MESOS_LOCAL_FILE_SERVER_PORT);
+    LOG.info("Using local port: " + port);
+    setLocalServerPort(Optional.fromNullable(port)); 
+  }
+  protected void setLocalServerPort(Optional<Integer> port) {
+      _localFileServerPort = port;
+  }
+  protected Optional<Integer> getLocalServerPort(){
```
erikdw: please add newlines between all of these methods

src/storm/mesos/MesosNimbus.java
```diff
@@ -97,57 +105,93 @@ public String getHostName(Map<String, SupervisorDetails> map, String nodeId) {
...
+      return _localFileServerPort;
+  }
+  protected void setScheduler(NimbusScheduler scheduler) {
+      _scheduler = scheduler;
+  }
+  protected NimbusScheduler getScheduler() {
+      return _scheduler;
+  }
+  protected void setupHttpServer() throws Exception {
```
erikdw: I like the refactoring in general, but could you please keep refactoring commits separate from the commit that addresses the bug/issue/task?

pdread100: To me they were one and the same but yes I will do that and keep the
refactoring to a minimum.

Thanks.

src/storm/mesos/MesosNimbus.java
pdread100: Yes I will do that.

Thanks

src/storm/mesos/MesosNimbus.java
pdread100: I will do that.

Thanks.

src/storm/mesos/MesosNimbus.java
pdread100: Generally its good practice to use getters/setters for instance variables,
especially public and protected. It encourages better unit testing,
extensibility, organization, and clarity. Its good OO as well.

Thanks

src/storm/mesos/MesosNimbus.java
pdread100: You do not like the @SuppressWarnings("unchecked") ??

In general I like to suppress warnings that are not going to get fixed
since the warning in that case adds nothing and its annoying from a
completeness stand point.

But I will remove it.

Thanks

pdread100: Yes I see I should have added the issues in my answers I was responding to,
I'm use to threads ....but I will fix all issues.

Thanks.

src/storm/mesos/MesosNimbus.java
pdread100: In src/storm/mesos/MesosNimbus.java
<https://github.com/mesos/storm/pull/37#discussion_r29535084>:

>      } catch (Exception e) {
>        throw new RuntimeException(e);
>      }
>    }
> -
> +
> +  @SuppressWarnings("unchecked")
> +  protected void initialize(Map conf, String localDir) throws Exception {
> +
> +      setConfig(conf);
> +      setLocalState(new LocalState(localDir));
> +
> +      _allowedHosts = listIntoSet((List<String>)conf.get(CONF_MESOS_ALLOWED_HOSTS));

>>> I wonder if listIntoSet can be replaced with google's gauve
Sets.newHashSet?


No, Sets.newHashSet throws a NullPointerException if the list is null.

Thanks

pdread100: >>> I wonder if listIntoSet can be replaced with google's gauve
Sets.newHashSet?

Here is a place where I would use a setter/getter. If you use

setAllowedHost (List<String> list) {

      if (list == null)

          _allowedHost = new HashSet<String>(0);

     else

         _allowedHost = new HashSet(list);

}


then in (Assume the same for disallowedHost)


public boolean isHostAccepted(String hostname) {
    return
        (_allowedHosts == null && _disallowedHosts == null) ||
            (_allowedHosts != null && _allowedHosts.contains(hostname)) ||
            (_disallowedHosts != null &&
!_disallowedHosts.contains(hostname))
        ;
  }


it would be

public boolean isHostAccepted(String hostname) {
    return (getAllowedHost().contains(hostname) ||
!getDisallowedHost().contains(hostname);
  }


The contract would hold up as long as the getters/setters were used.


Thoughts?

src/storm/mesos/MesosNimbus.java
pdread100: What if we call it getCredentials() instead of getAuthorization()? It would seem more appropriate.

src/storm/mesos/MesosNimbus.java
erikdw: @pdread100 : you misunderstood Tim.  He meant "newline" rather than "new line". ;-)   The annotation is fine.

pdread100: Ok. Thanks for the clarification.

src/storm/mesos/MesosNimbus.java
erikdw: Can you annotate your comments with markdown?    https://help.github.com/articles/github-flavored-markdown/

i.e., put ``` above and below your code blocks.   And don't wrap them unnecessarily (I'm like the protagonist from "the princess and the pea" -- I cannot read code that is malformed).

src/storm/mesos/MesosNimbus.java
erikdw: Well, `getAuthorization()` is different from `getAuthentication()`   (http://stackoverflow.com/questions/6556522/authentication-versus-authorization)  :-)

But yes, `getCredentials()` sounds good!

src/storm/mesos/MesosNimbus.java
pdread100: Certainly...I'm a little new to this git stuff...still feeling my way.

Thanks.

erikdw: Note that you can edit your existing "line notes" by clicking the pencil icon above them.

src/storm/mesos/MesosNimbus.java
tnachen: is anyone else calling these setters though? I don't really think we should bloat the class unless we have someone actually calling these. We should also make class variables private if no one else is setting them.

src/storm/mesos/MesosNimbus.java
tnachen: @pdread100 I think your refactoring is reasonable, but then it makes this commit a lot harder to merge now since there are a lot of changes and also some changes are related to a new feature and mostly are just refactoring.

src/storm/mesos/MesosNimbus.java
erikdw: +1

src/storm/mesos/MesosNimbus.java
```diff
@@ -707,7 +751,58 @@ public void executorLost(SchedulerDriver driver, ExecutorID executor, SlaveID sl
...
+        String principal = Optional.fromNullable((String)getConfig().get(CONF_MESOS_PRINCIPAL)).orNull();
+        String secretFilename = Optional.fromNullable((String)getConfig().get(CONF_MESOS_SECRET_FILE)).orNull();;
+        
+        if (principal != null) {         
+            finfo.setPrincipal(principal);
+            Credential.Builder credentialBuilder = Credential.newBuilder();
+            credentialBuilder.setPrincipal(principal);
+            if (StringUtils.isNotEmpty(secretFilename)) {
+                try {
+                    // The secret cannot have a NL after it
```
tnachen: let's spell out NL

pdread100:
I have made all the changes requested by @tnachen. Is this sufficient? Thanks.

erikdw:
hey @pdread100, irrespective of the outcome of @tnachen's response, can you please squash your commits into a single commit, and then force push?  So that we get a clean single commit from this change.  Thanks!   If you need guidance on doing that let me know.

erikdw:
@pdread100 : I just took a gander at your changes -- it seems that you missed @tnachen's request to not add all these getter/setter methods for variables which are only being used internally to the class.  I agree with Tim, it's bloating the class unnecessarily.

tnachen:
@erikdw I think I can squash them myself and remove the getter/setter since this has been sitting for a while. @pdread100 the changes looks good to me, but can you rebase on master? I cannot apply this PR on latest master.

erikdw:
@tnachen : great, thanks!

pdread100:
The last two days I have not had access to the net nor can I address this
until monday. Sorry for the not covering the issues but I thought the
setter/getter thing was not really an issue. I have been trying to squash
the commits but have not been too succesful.

Thanks

pdread100:
Ok. Rebased, squashed commits, removed the "protected" setters/getters (they were not private), should be good to go.

src/storm/mesos/MesosNimbus.java
```diff
@@ -20,10 +20,12 @@
 import backtype.storm.Config;
 import backtype.storm.scheduler.*;
 import backtype.storm.utils.LocalState;
+
```
brndnmtthws: Whitespace.

src/storm/mesos/MesosNimbus.java
```diff
@@ -105,54 +112,65 @@ public String getHostName(Map<String, SupervisorDetails> map, String nodeId) {
...
-      }
-
-      Integer port = (Integer) _conf.get(CONF_MESOS_LOCAL_FILE_SERVER_PORT);
-      _localFileServerPort = Optional.fromNullable(port);
-      LOG.debug("LocalFileServer configured to listen on port: " + port);
-
+  protected void setupHttpServer() throws Exception {
       _httpServer = new LocalFileServer();
       _configUrl = _httpServer.serveDir("/conf", "conf", _localFileServerPort);
+
```
brndnmtthws: Whitespace.

brndnmtthws:
Open a separate PR for the whitespace/formatting changes.

maverick2202:
What username will the storm mesos framework use to run the storm topology ? Default, i think is root. Can we add it as a parameter too ?

pdread100:
Mesos ACL will determine what the user tasks/topology is run as.

See Mesos Autorization
<https://mesos.apache.org/documentation/latest/authorization/>

maverick2202:
Mesos ACL defines rules on how to run the task. The user account comes from the framework. In storm mesos framework, it is not set. 

FrameworkInfo.Builder finfo = FrameworkInfo.newBuilder()
.........
.setUser("")

 Can we also make it as a parameter ?

pdread100:
I will double check but I believe you are incorrect. I believe mesos/myriad
is set up the exact same way an I think they do not set the user with the
explanation that mesos will handle adding the user.

erikdw:
The formatting is all totally out-of-whack after this change.  The new code was written with 4-space indents, the old code was 2-space indents, and some of the functions run into each other (no newline in-between).   I'll do a reformatting PR.

pdread100:
I think we need a style checker, similar to what mesos/myriad uses. Such
that once code passes the style checker it should be good to go. I will
create a PR with an example one which can be tweaked to the satisfaction of
the group.

pdread100:
I have created an issue to be worked in response to your request, issue
<https://github.com/mesos/storm/issues/43>

erikdw:
Style checkers are better than nothing. Barely.  They cannot / don't handle a lot of stuff, I personally hate them.

- Erik

pdread100:
Yeah I'm not a fan either, however I have just started working with open
source projects and I'm running into the style buzz saw, some is legit and
a lot just absurd but it has to pass muster to get it into the repo.....
and it would be nice if there was some consistency from project to project.

I just ran two different style checkers and this project would require
massive "fixes" to get it to the level of other projects. And thats not
counting indents which is another sore spot. I simply don't understand how
anyone can justify 2 space indents...why even bother trying to make
something readable. Just make it straight line code and be done with it.
And the fear of a NL, holy cow bat man!

Anyway I will push a PR and folks can test it and if its not a good fit for
this project so be it.

erikdw:
Sounds great, thanks!

- Erik

Issue #6, opened 2014-08-13
Logviewer ?


yaronr:
Hi

I would like to access the storm logviewer on one of the supervisors.
How do I do that?

(AFAIK the supervisors are started by mesos)

Thanks

SEJeff:
@yaronr give me about a week to finish up and I'll post how I got it all running ontop of mesos + marathon without hardcoding the nimbus master or anything.

The gist of it is:

1. A marathon [application group](https://mesosphere.github.io/marathon/docs/application-groups.html). One for nimbus, and one for the storm ui. The app for storm ui has a dependency on nimbus, so one starts after the other.
2. A run-storm.py which stores the actual host and port of the nimbus instance in zookeeper using the excellent [Kazoo](https://kazoo.readthedocs.org/en/latest/) python lib.
3. run-storm.py then writes out the `conf/storm.yaml` with the proper `nimbus.thrift.port` an `nimbus.host` settings before launching nimbus
4. It then writes out `ui.port`, `nimbus.host`, and `nimbus.thrift.port` when launching storm ui.

All of this and I actually don't run it in docker, but natively under marathon, by downloading the built storm distribution and the run-storm.py via the `uris` flag in marathon json. I'm also chrooting this in a specific zookeeper path and having the framework name include that chroot path, so that multiple storm frameworks can run, potentially as different users, on a single mesos cluster.

In a perfect world, the framework would support a lot of this natively, but I'm a python/golang/c guy, and have never taken a serious stab at java.

WalterDalton:
SEJeff - Could we use something like mesos-dns to indicate where the nimbus is located? It works well for Myriad and may work well here.

SEJeff:
It kind of defeats the point of mesos to require hard coding masters that might change. Mesos DNS is a reasonable workaround, as is my solution of having a python script write to ZK before starting nimbus and another to read from nimbus before starting storm ui.

The framework should do this natively.

WalterDalton:
Makes sense that the framework should do it natively, I was just suggestion the mesos-dns as a way that the nimbus/ui could be run from any node (given access to the code to run it) and where it landed wouldn't matter to the supervisors as they are pointing to a static name that resolves to a dynamic IP.

SEJeff:
True. Thanks for the suggestion

pdread100:
Wouldn't the most straight forward approach be to have the supervisor/executor fire up the logviewer when mesos fires it up? 

If no one is working this I would like to get the logview working under mesos.

[...]

PR #38, opened 2015-05-19
Added logviewer startup from supervisor

pdread100:
This launches a logviewer by default when a supervisor is launched by mesos. There is a new flag supervisor.autostart.logviewer which if set to false will prevent the logviewer from being launched. This was tested on a Vagrant 2 node cluster. This addresses issue [#6] (https://github.com/mesos/storm/issues/6)

I have the Vagrant setup/configuration if anyone would like PR created for that.

### DIFF
--- a/bin/build-release.sh
+++ b/bin/build-release.sh
@@ -42,8 +42,8 @@ function clean {
 }
 
 function mvnPackage {
-  mvn -Dcheckstyle.skip=true package
-  mvn -Dcheckstyle.skip=true dependency:copy-dependencies
+  mvn package
+  mvn dependency:copy-dependencies
 }
 
 function prePackage {(
--- a/src/storm/mesos/MesosCommon.java
+++ b/src/storm/mesos/MesosCommon.java
@@ -30,6 +30,7 @@ public class MesosCommon {
   public static final String EXECUTOR_CPU_CONF = "topology.mesos.executor.cpu";
   public static final String EXECUTOR_MEM_CONF = "topology.mesos.executor.mem.mb";
   public static final String SUICIDE_CONF = "mesos.supervisor.suicide.inactive.timeout.secs";
+  public static final String AUTO_START_LOGVIEWER_CONF = "supervisor.autostart.logviewer";
 
   public static final double DEFAULT_CPU = 1;
   public static final double DEFAULT_MEM_MB = 1000;
--- a/src/storm/mesos/MesosSupervisor.java
+++ b/src/storm/mesos/MesosSupervisor.java
@@ -27,6 +27,11 @@ import org.apache.mesos.MesosExecutorDriver;
 import org.apache.mesos.Protos.*;
 import org.json.simple.JSONValue;
 
+import com.google.common.base.Optional;
+
+import storm.mesos.logviewer.ILogController;
+import storm.mesos.logviewer.LogViewerController;
+
 import java.io.IOException;
 import java.util.Collection;
 import java.util.HashSet;
@@ -69,6 +74,13 @@ public class MesosSupervisor implements ISupervisor {
     LOG.info("Waiting for executor to initialize...");
     try {
       _executor.waitUntilRegistered();
+      
+      if (startLogViewer(conf)) {
+        LOG.info("Starting logviewer...");
+        ILogController logController = new LogViewerController();
+        logController.start();
+      }
+      
     } catch (InterruptedException e) {
       throw new RuntimeException(e);
     }
@@ -182,6 +194,10 @@ public class MesosSupervisor implements ISupervisor {
 
   }
 
+  protected boolean startLogViewer(Map conf) {
+     return Optional.fromNullable((Boolean) conf.get(MesosCommon.AUTO_START_LOGVIEWER_CONF)).or(true);
+  }
+  
   public class SuicideDetector extends Thread {
     long _lastTime = System.currentTimeMillis();
     int _timeoutSecs;
--- a/src/storm/mesos/logviewer/LogViewerController.java
+++ b/src/storm/mesos/logviewer/LogViewerController.java
@@ -1,0 +1,169 @@
+package storm.mesos.logviewer;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.lang.ProcessBuilder.Redirect;
+import java.nio.file.Paths;
+import java.util.Map;
+
+import org.apache.commons.io.IOUtils;
+import org.apache.log4j.Logger;
+import org.yaml.snakeyaml.Yaml;
+
+import com.google.common.base.Optional;
+
+public class LogViewerController implements ILogController {
+    private static final Logger LOG = Logger.getLogger(LogViewerController.class);
+    protected Process process;
+    protected IUrlDetection urlDetector;
+    protected String installLocation;
+    protected Map<String, Object> config;
+    protected Optional<Integer> port;
+    protected static final String LOG_VIEWER_PORT = "logviewer.port";
+    
+    /**
+     * Create a log view controller with a SocketUrlDetector set to 
+     * the port that is in the storm.yaml config file. If no port is
+     * set then use the default of 8000. 
+     */
+    public LogViewerController() {
+        // Use a simplistic approach.
+        this(new SocketUrlDetection());
+    }
+    
+    /**
+     * Load a preconfigured Url detector to check for an existing logviewer.
+     * 
+     * @param urlDetector
+     */
+    public LogViewerController(IUrlDetection urlDetector) {
+        setUrlDetector(urlDetector);
+        initialize();
+    }
+    
+    /**
+     * Start up the logviewer, but before that is done a check is made to
+     * see if an existing logviewer (or process) is on the same port and report
+     * and error if so.
+     */
+    @Override
+    public void start() {
+       try {
+           if (!exists()){
+               launchLogViewer();
+           } else {
+               LOG.error("Failed to start logviewer because there is something on its port");
+           }    
+       } catch (Exception e) {
+           LOG.error("Failed to start logviewer", e);
+       }
+    }
+    
+    @Override
+    public void stop() {
+        getProcess().destroy();
+    }
+    
+    @Override
+    public boolean exists() {
+        return getUrlDetector().isReachable();
+    }
+    
+    public void setUrlDetector(IUrlDetection urlDetector) {
+        this.urlDetector = urlDetector;
+    }
+    
+    public IUrlDetection getUrlDetector() {
+        return urlDetector;
+    }
+    
+    public void setInstallLocation(String location) {
+        installLocation = location;
+    }
+    
+    public String getInstallLocation() {
+        return installLocation;
+    }
+
+    protected void setProcess(Process process) {
+        this.process = process;
+    }
+    
+    protected Process getProcess() {
+        return process;
+    }
+    
+    protected void setPort(Optional<Integer> oport) {
+        port = oport;
+    }
+    
+    protected Optional<Integer> getPort() {
+        return port;
+    }
+    
+    protected void setConfig(Map<String, Object> configuration) {
+        config = configuration;
+    }
+    
+    protected Map<String, Object> getConfig(){
+        return config;
+    }
+
+    protected void launchLogViewer () throws IOException {
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("launchLogViewer: home dir? " + getInstallLocation());
+        }
+        ProcessBuilder pb = createProcessBuilder(getInstallLocation());
+        setProcess(pb.start());       
+    }
+    
+    /**
+     * Create a process builder to launch the log viewer
+     * @param logDirectory
+     * @return
+     */
+    protected ProcessBuilder createProcessBuilder(String homeDirectory) {
+        ProcessBuilder pb = new ProcessBuilder(Paths.get(homeDirectory, "/bin/storm").toString(), "logviewer");
+         
+        // If anything goes wrong at startup we want to see it.
+        File log = Paths.get(homeDirectory, "/logs/logviewer-startup.log").toFile();
+        pb.redirectErrorStream(true);
+        pb.redirectOutput(Redirect.appendTo(log)); 
+        return pb;
+    }
+    
+    @SuppressWarnings("unchecked")
+    protected void initialize() {
+        setInstallLocation(new File(".").getAbsolutePath());
+        
+        loadConfig();
+        
+        // If the storm configuration has defined a port use it.
+        if (getPort().isPresent()) {
+            getUrlDetector().setPort(getPort());
+        } else {
+            // Otherwise use the default 
+            LOG.info("No valid port in storm.yaml, using the on configured in the Url Detector " + 
+                getUrlDetector().getPort());
+            setPort(getUrlDetector().getPort());
+        }
+    }
+    
+    @SuppressWarnings("unchecked")
+    protected void loadConfig() {
+        InputStream input = null;
+        try {
+            input = new FileInputStream(Paths.get(getInstallLocation(), "/conf/storm.yaml").toFile());
+            setConfig((Map<String, Object>) new Yaml().load(input));
+            LOG.info("Setting port for log viewer to " + config.get(LOG_VIEWER_PORT));
+            setPort(Optional.of((Integer) getConfig().get(LOG_VIEWER_PORT)));
+        } catch (Exception e) {
+            LOG.error("Failed to configure the log viewer controller,", e);
+            setPort(Optional.fromNullable(((Integer) null)));
+        } finally {
+            IOUtils.closeQuietly(input);
+        }
+    }
+}
--- a/src/storm/mesos/logviewer/SocketUrlDetection.java
+++ b/src/storm/mesos/logviewer/SocketUrlDetection.java
@@ -1,0 +1,59 @@
+package storm.mesos.logviewer;
+
+import java.net.InetAddress;
+import java.net.Socket;
+
+import org.apache.commons.io.IOUtils;
+import org.apache.log4j.Logger;
+
+import com.google.common.base.Optional;
+
+public class SocketUrlDetection implements IUrlDetection {
+    public static final int DEFAULT_PORT = 8000;
+    protected Optional<Integer> port;
+    private static final Logger LOG = Logger.getLogger(SocketUrlDetection.class);
+    
+    public SocketUrlDetection() {
+        setPort(Optional.of(DEFAULT_PORT));
+    }
+    
+    public SocketUrlDetection(int port) {
+        setPort(Optional.of(port));
+    }
+    
+    public SocketUrlDetection(Optional<Integer> port) {
+        setPort(port);
+    }
+    
+    @Override
+    public boolean isReachable() {
+        Socket socket = null;
+        boolean reachable = false;
+        try {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Checking host " + InetAddress.getLocalHost() + " and port " + getPort());
+            }
+            
+            if (port.isPresent()) {
+                socket = new Socket(InetAddress.getLocalHost(), getPort().get());
+                reachable = true;
+            }
+        } catch (Exception e) {
+            // don't care.
+        } finally {
+            IOUtils.closeQuietly(socket);
+        }
+        return reachable;
+    }
+    
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    @Override
+    public void setPort(Optional port) {
+        this.port = (Optional<Integer>) port;
+    }
+    
+    @Override
+    public Optional<Integer> getPort() {
+        return port;
+    }
+}
--- a/README.markdown
+++ b/README.markdown
@@ -109,7 +109,7 @@ Storm/Mesos provides resource isolation between topologies. So you don't need to
 * `mesos.framework.name`: Framework name. Defaults to "Storm!!!".
 * `mesos.framework.principal`: Framework principal to use to register with Mesos
 * `mesos.framework.secret.file`:  Location of file that contains the principal's secret. Secret cannot end with a NL.
-
+* `supervisor.autostart.logviewer`: Default is true, if not false please add 128M to topology.mesos.executor.mem.mb
 
 ## Resource configuration
 
--- a/pom.xml
+++ b/pom.xml
@@ -155,5 +155,10 @@
         <artifactId>guava</artifactId>
         <version>18.0</version>
       </dependency>
+      <dependency>
+	<groupId>junit</groupId>
+	<artifactId>junit</artifactId>
+	<version>4.8.1</version>
+      </dependency>
     </dependencies>
   </project>
--- a/src/storm/mesos/logviewer/ILogController.java
+++ b/src/storm/mesos/logviewer/ILogController.java
@@ -1,0 +1,11 @@
+package storm.mesos.logviewer;
+
+public interface ILogController {
+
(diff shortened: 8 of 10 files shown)


src/storm/mesos/MesosCommon.java
```diff
@@ -18,6 +18,7 @@
 package storm.mesos;
 
 import backtype.storm.scheduler.TopologyDetails;
+
```
brndnmtthws: Whitespace.

src/storm/mesos/MesosCommon.java
```diff
@@ -30,8 +31,10 @@
   public static final String EXECUTOR_CPU_CONF = "topology.mesos.executor.cpu";
   public static final String EXECUTOR_MEM_CONF = "topology.mesos.executor.mem.mb";
   public static final String SUICIDE_CONF = "mesos.supervisor.suicide.inactive.timeout.secs";
+  public static final String AUTO_START_LOGVIEWER_CONF = "supervisor.autostart.logviewer";
 
   public static final double DEFAULT_CPU = 1;
+
```
brndnmtthws: Whitespace.

src/storm/mesos/MesosSupervisor.java
```diff
@@ -20,6 +20,7 @@
 import backtype.storm.scheduler.ISupervisor;
 import backtype.storm.utils.LocalState;
 import backtype.storm.utils.Utils;
+
```
brndnmtthws: Whitespace.

src/storm/mesos/logviewer/LogViewerController.java
```diff
@@ -0,0 +1,210 @@
...
+    /**
+     * Create a process builder to launch the log viewer
+     * @param logDirectory
+     * @return
+     */
+    protected ProcessBuilder createProcessBuilder(String homeDirectory) {
+        ProcessBuilder pb = new ProcessBuilder(homeDirectory + "/bin/storm","logviewer");
+         
+        // If anything goes wrong at startup we want to see it.
+        File log = new File(homeDirectory + "/logs/logviewer-startup.log");
```
brndnmtthws: Is there some `Path.join()` thing you can use instead? Something more idiomatic.

src/storm/mesos/logviewer/LogViewerController.java
```diff
@@ -0,0 +1,210 @@
...
+        pb.redirectErrorStream(true);
+        pb.redirectOutput(Redirect.appendTo(log)); 
+        return pb;
+    }
+    
+    @SuppressWarnings("unchecked")
+    protected void initialize() {
+        setInstallLocation(new File("./").getAbsolutePath());
```
brndnmtthws: `new File("./").getAbsolutePath()` seems like a weird way to get the current working dir. Can you do this in a more idiomatic way?

src/storm/mesos/logviewer/LogViewerController.java
```diff
@@ -0,0 +1,210 @@
...
+                getUrlDetector().getPort());
+            setPort(getUrlDetector().getPort());
+        }        
+    }
+    
+    @SuppressWarnings("unchecked")
+    protected void loadConfig() {
+        InputStream input = null;
+        try {
+            input = new FileInputStream(new File(getInstallLocation() + "/conf/storm.yaml"));
```
brndnmtthws: Again, maybe join the paths the idiomatic way. I think you can do `new File(path1, path2)` in Java.

src/storm/mesos/logviewer/SocketUrlDetection.java
```diff
@@ -0,0 +1,66 @@
...
+    
+    @Override
+    public boolean isReachable() {
+
+        Socket socket       = null;
+        boolean reachable   = false;
+        try {
+            
+            if (LOG.isDebugEnabled())
+                LOG.debug("Checking host "+ InetAddress.getLocalHost() + "and port " + getPort() );
```
brndnmtthws: Space between `"+`.

brndnmtthws: And also between `"and port "`.

brndnmtthws:
Made a few comments, mostly related to formatting. Otherwise looks good.

brndnmtthws:
For the whitespace/formatting changes, open a separate PR if you want to make those.

brndnmtthws:
Also, could you please update the docs or make a note about allocating additional memory overhead for the container to account for the logviewer JVM.

pdread100:
I fixed all issues except the getting the local installation directory. I'm betting mesos in some way informs the supervisor of the installation directory but I have not been able to find how. That can be addressed later. Since I had to rebase because of the merge I simply squashed the commits and pushed instead of issuing a new PR. If thats not satisfactory my apologies.

src/storm/mesos/logviewer/LogViewerController.java
tnachen: Even if you want to do it this way, "./" seems to already assume this is linux only. I think you want to just do File(".") which should give you the same result?

pdread100: I think you are correct and I will change it. Thanks.

src/storm/mesos/logviewer/LogViewerController.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+    
+    @Override
+    public boolean exists() {
+        return getUrlDetector().isReachable();
+    }
+    
+    @Override
+    public void terminateExistingViewer() {
+        LOG.info("Terminating existing viewer....");
```
tnachen: I don't quite understand, are you hoping to merge this as is? This doesn't look like it will ever work ?

pdread100: I felt there should be an effort to check if the logviewer, a logviewer,
was up and make an attempt to in some manner kill it. I have not come up
with a better way of dealing with it but some one may have a better idea.
The only thing that is realistic is leverage requirements against the mesos
logviewer to add shutdown rest service, or if the python code could
generate a PID (I'm not a python person) and do a kill on that but I felt
the algorithm should be there.

The current behavior is it dies when the supervisor dies.

If you would prefer the algorithm not be there thats fine but I would
expect at least comments should be there to address the issue.

pdread100:
@tnachen @brndnmtthws I made the changes to logviewer, also the other PRs...could be we continue the dialog/process? 
Thanks.

src/storm/mesos/MesosSupervisor.java
```diff
@@ -182,6 +194,10 @@ public void disconnected(ExecutorDriver driver) {
 
   }
 
+  protected boolean startLogViewer(Map conf) {
+     return Optional.fromNullable((Boolean) conf.get(MesosCommon.AUTO_START_LOGVIEWER_CONF)).or(true);
```
tnachen: I think in the future we should make some helpers to automatically get default value

src/storm/mesos/logviewer/SocketUrlDetection.java
```diff
@@ -0,0 +1,66 @@
...
+        setPort(Optional.of(port));
+    }
+    
+    public SocketUrlDetection(Optional<Integer> port) {
+        setPort(port);
+    }
+    
+    @Override
+    public boolean isReachable() {
+
```
tnachen: Remove extra space

src/storm/mesos/logviewer/SocketUrlDetection.java
```diff
@@ -0,0 +1,66 @@
...
+        Socket socket       = null;
```
tnachen: I dont' think we want to encourage this alignment in the code, can you simply just use the regular spacing?

src/storm/mesos/logviewer/SocketUrlDetection.java
```diff
@@ -0,0 +1,66 @@
...
+        boolean reachable   = false;
+        try {
+            
+            if (LOG.isDebugEnabled())
+                LOG.debug("Checking host " + InetAddress.getLocalHost() + " and port " + getPort() );
```
tnachen: Remove extra space in the end

src/storm/mesos/logviewer/SocketUrlDetection.java
```diff
@@ -0,0 +1,66 @@
...
+            
+            if (port.isPresent()) {
+                socket      = new Socket(InetAddress.getLocalHost(), getPort().get());
```
tnachen: Spacing

tnachen: You'll need to close this socket in finally?

src/storm/mesos/logviewer/SocketUrlDetection.java
```diff
@@ -0,0 +1,66 @@
...
+        }
+        finally {
+            IOUtils.closeQuietly(socket);
+        }
+        return reachable;
+    }
+    
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    @Override
+    public void setPort(Optional port) {
```
tnachen: Why not Optional<Integer>?

tnachen: Actually looking at the code I don't even know what's the point of the optional, we always set a default and then we never allow it to be empty, so why not just a int here

src/storm/mesos/logviewer/SocketUrlDetection.java
```diff
@@ -0,0 +1,66 @@
...
+    } 
+    
+    @Override
+    public Optional<Integer> getPort() {
+        return port;
+    }
+    
+    protected Optional<Integer> port;
+ 
+    private static final Logger LOG = Logger.getLogger(SocketUrlDetection.class);
```
tnachen: Variables are on the top of the class

src/storm/mesos/logviewer/SocketUrlDetection.java
pdread100: I guess I will never understand the phobia with spacing and newlines. I
have always felt it added readability to the code....but if I must I will
remove them.

src/storm/mesos/logviewer/LogViewerController.java
```diff
@@ -0,0 +1,180 @@
...
+    protected void loadConfig() {
+        InputStream input = null;
+        try {
+            input = new FileInputStream(Paths.get(getInstallLocation(),"/conf/storm.yaml").toFile());
+            setConfig((Map<String, Object>)new Yaml().load(input));
+            LOG.info("Setting port for log viewer to " + config.get(LOG_VIEWER_PORT));
+            setPort(Optional.of((Integer)getConfig().get(LOG_VIEWER_PORT)));
+        } catch (Exception e) {
+            LOG.error("Failed to configure the log viewer controller,", e);
+            setPort(Optional.fromNullable(((Integer)null)));
```
tnachen: You think setPort will throw a exception here? That's the only motivation I Can think of for you to set this back to null.

tnachen:
@brndnmtthws @erikdw Any comments on how the logviewer is being launched? Let's try to help @pdread100 to get this merged soon

src/storm/mesos/logviewer/SocketUrlDetection.java
pdread100: Yep, the incorrect way of handling private variables which should always be
at the bottom. Software engineers are lemmings by nature, thats how the
whole privates at the top got started..... but if I must.

src/storm/mesos/logviewer/LogViewerController.java
pdread100: This was actually simply handling the exception thrown by FileInputStream. Setting the port is more to insure the Optional was set correctly, not so much the "port" was set. The Optional thing was because folks here love Optionals.....I'm not a big fan and will happily remove them.

src/storm/mesos/logviewer/SocketUrlDetection.java
tnachen: I think it's more about being consistent with other code in the same repo. So unless you change all of them and put it in the code style, or stick with what's around and propose something later.

src/storm/mesos/logviewer/SocketUrlDetection.java
pdread100: The reason its and Optional and should stay that way is because the IUrlDetection interface requires an Optional. And as previously stated the one implementation of that interface uses an Optional. It sort of all is integrated and plays together. At this point I feel it should stay as is.

pdread100:
@tnachen @brndnmtthws @erikdw All style fixes are in place. Optionals I felt where needed per comments. Thanks for your help!

src/storm/mesos/logviewer/SocketUrlDetection.java
tnachen: hmm I see, it's fine let's leave it as is then.

tnachen:
I'm going to merge this then since there is no more comments from anyone.

erikdw:
@pdread100 & @tnachen & @brndnmtthws : sorry for the late reply, I've been busy with work (and didn't even notice this functional change was out for review until @tnachen copied me 2 days ago -- was doing [late binding](http://www.jnd.org/dn.mss/why_procrastination_.html) on the checkstyle stuff).

So I hope you won't mind bearing with me as I ask for some clarification on how this works regarding the reading of the storm worker logs.  Since the worker logs are in executor/topology-specific dirs, how does a single logviewer on a host know where to find all the different logs?

erikdw:
Also, @pdread100 : w.r.t. your initial commit's statement:  "I have the Vagrant setup/configuration if anyone would like PR created for that." --> I would really appreciate seeing what it looks like please.  Maybe you can direct mail me a link to a branch?

bin/build-release.sh
```diff
@@ -42,8 +42,8 @@ function clean {
 }
 
 function mvnPackage {
-  mvn -Dcheckstyle.skip=true package
-  mvn -Dcheckstyle.skip=true dependency:copy-dependencies
+  mvn package
```
erikdw: did we really wanna disable the running of checkstyle here?

pdread100: No. This should have been corrected. I will see what happened since my
feature branch which this PR is based on has it removed.

pdread100: The merged buiild-release.sh has the correct setting. That is to say the
style check is run by default.

pdread100:
I already create the PR for the Vagrant cluster setup and it has been
merged. If you look at

https://github.com/mesos/storm/blob/master/docs/vagrant.md

You will see how to set it up.

pdread100:
Actually there is a logviewer on every executor host and I will double
check but I believe the Storm UI feeds the log URL to the logviewer on the
specific host. I ran all the sample topos on two nodes and it had no
problem finding them.

Also in general the only difference from the normal storm logviewer is the
mesos storm executor is launching the storm logviewer using the normal
storm mechanism which should function as if it were manually launched from
the command line. In addition we account for the memory foot print in the
offers.

I would think the storm UI should be launched in a similar manner for the
same reasons. Maybe even incorporated into the scheduler directly like
mesos/myriad does theirs. That way its full mesos driven and all the
memory/cpu accounting is handled by mesos.

erikdw:
@pdread100 : thanks for the responses!  

Regarding the single-logviewer, I still have some doubts.

Have you tried having multiple separate MesosSupervisors (mesos-executors) on the same mesos-slave host (i.e., 2 topologies having storm workers scheduled into the same mesos-slave host)?

The way it works is that each mesos executor has a separate working dir for its encapsulated application.  So each separate MesosSupervisor on a single host will have a separate location for its worker logs, if I'm not mistaken.   If you only run a single logviewer, then I'm not seeing how it can serve up all the worker logs from the separate executors.

Furthermore, since this logviewer seems to be launched within the container of the MesosSupervisor, if that supervisor's topology's workers on this host die, then the container will get cleaned up, and it will nuke the logviewer.

erikdw:
Just FYI, in my company's setup, we have hacked storm to log to a central shared dir outside of the container, but that's not the standard setup for using storm in mesos (AFIK).

Oh, one other thing -- I believe that you're replying via email, so I'm guessing that's why  you're missing updates I'm making to the comments I've left. So 2 of your replies were for questions I had which I already answered myself and then I had updated the comments.  :)  (i.e., my misreading of the checkstyle change & my missing of the vagrant PR)

pdread100:
It sounds like this is a bigger issue than this PR addresses. It may be a
storm issue and not a mesos/storm issue. You may want to create the issue
to get this dialog to a larger audience. I need to do more testing but its
going to be a while before I get back to this .

Thanks for the heads up on the messages. I have a bad habit of responding
via email since its much faster than returning to the PR. But I need to
break that bad habit.

erikdw:
@pdread100 : I'd probably classify it as a storm-on-mesos issue, rather than either a mesos or a storm issue.  I feel like we have 3 options:

1. Don't bother with logviewer at all (probably not best solution long-term, maybe best for *right now* -- implies backing out this change to avoid unnecessary complications in the code)
2. Launch logviewers within each MesosSupervisor executor's container.  i.e., basically what you've done here, but removing the "don't launch more than 1 logviewer per host" logic, and changing the listen port be 0 (thus causing the jetty listen socket creation to get an ephemeral port), then _somehow_ register the obtained port with the nimbus for the set of worker ports owned by this MesosSupervisor.  Potentially hard to keep in-sync as worker processes come-and-go on a host.
3. Launch a separate mesos-executor on each host for the logviewer, and somehow allow it to serve logs from multiple log dirs.
4. Some combination of the above, with a distributed registration solution involving mesos-dns (or something like that) for registering logviewers with the nimbus.  (@SEJeff was talking about something like this on https://github.com/mesos/storm/issues/6 ).

Notably, all of this seems like it will require changes to storm.

I'm feeling like it might be best to go with 1. right now and back out this change until we decide how to more properly handle logviewer being launched?   @tnachen & @brndnmtthws any thoughts about that proposal?

pdread100:
@erikdw @tnachen @brndnmtthws For now I would rather this pull request was not backed out. My customer only runs one executor per host and really wants this feature. So until a multiple executor per host logviewer is ready would it hurt to leave this in place?

tnachen:
Hi @erikdw thanks for chiming in, from your options to me it sounds like 3 seems to be the cleanest impl to support multiple log dirs right?
How hard is it to do that?

erikdw:
@tnachen : to implement would probably require changing storm and getting it released -- i.e., it will likely _never_ work with the version of storm we currently build mesos-storm against (0.9.3).  Nor the next 2 which have already been released: 0.9.4, 0.9.5.    Maybe can squeeze it into 0.10.0, that's gone to beta.  I'm not sure what their release process is like for whether it would.  Complications: the logviewer needs to be updated for each additional mesos-executor/MesosSupervisor/topology that arises on this host.

@pdread100 : so your customer is only using 1 topology?  (If they used more than 1, and if there were enough resources on each host to hold more than 1 topology, then there could be 2+ mesos-executors on each host).  I don't believe that's a typical deployment for storm.   I definitely would want this off by default if we leave the logic in place.

pdread100:
@erikdw They are definitely going to want the capability to handle multiple executors, log dirs, but this could take a while to put in and I know they want to exercise the current implementation if possible. But if push comes to shove and its backed then so be it. Since mesos/storm is pretty new I suspect their view of the world will change fairly rapidly. I personally just deal with their current requirements and they also set the priority so that could change in the next 15 min....so like I said if its going to take 6 months to get the new version in then it would be nice to have it the way it is...if its going to take < 45 days then no big deal .... As you pointed out the real solution is going to be with modifying the Storm logviewer... either adding a rest service that some process (executor/nimbus) can push new config info to it or a zookeeper structure update.... so let the chips fall where they may.

tnachen:
@erikdw @pdread100 so how about let's disable this by default and also add some comments this only works for one topology then? And are you guys going to make changes to storm so it can support multiple logs dir?

erikdw:
@tnachen : sounds like that will work for now, though I'm a bit skeeved out by the high likelihood of leaving this only partially functional change in there for a long time.

I can file a bug in storm's apache repo, but cannot promise that I'll work on it.  This is a non-trivial change, beyond the "need to keep up-to-date with set of mesos-executors on a host" problem, the logviewer isn't written to simply take a raw dir.  It extract the "log dir" from the logback config on the storm worker host:
* https://github.com/apache/storm/blob/0c2b3a4f640d1b2905c093eb85f03190ceb0a711/storm-core/src/clj/backtype/storm/daemon/logviewer.clj#L208-217

A proper solution for this overall problem requires more thought and design.

erikdw:
@tnachen & @pdread100 : after pondering over this in the back of my mind, I wonder if we can take a simpler approach:  have the storm workers log to a common dir outside of the mesos-executor (i.e., how my company has set up storm's logback config), then we can just launch 1 logviewer per host (basically only a slight change beyond #38 ) and then we don't need to change storm and introduce lots of complications.   This also avoids problems with the logs getting garbage-collected when the mesos supervisor goes away.   I'm not sure if this runs contrary to the general mesos philosophy though.

pdread100:
@erikdw @tnachen If using a common dir is fine with mesos/storm then thats fine with me. Erik would you mind giving me an idea on how that should be approached? If you have already done it then no point in me re-inventing the wheel. Also I'm not sure if it matters but the current implementation allows for disabling the logviewer from being launched. 

Also we need to provide for dynamic port allocation...not sure how thats going to happen yet but thats being discussed locally.

pdread100:
@erikdw How are you creating the second instance of the supervisor? Testing and what I can find on the net indicate only one supervisor starts regardless of the concurrent topologies. So you must be starting it manually, is this correct? Something like bin/storm-mesos supervisor &  ??

erikdw:
@pdread100 : 

1. Regarding how to have a common logdir:  we need to pass the logstash config into the MesosSupervisor processes, and have that point to the common dir.  I need to figure out how running storm "normally" in the mesos/storm framework really works to be more specific (at my company we have an unfortunate hack I inherited for running mesos/storm in our environment).

2. Regarding having more than 1 supervisor launch:  you just need to have more than 1 topology get scheduled onto a single host.  Each storm supervisor is a MesosSupervisor which is a mesos-executor.  Here's one example for how *you* can do it with your vagrant setup (#48):
* submit 2 separate topologies
* each topology should have these settings:
 * `topology.workers`: 1
 * `topology.mesos.worker.cpu`: 1
 * `topology.mesos.worker.mem.mb`: 1024

That should cause 2 topologies to be running on a single worker/slave host, which will cause 2 supervisors to be launched.  The reason separate mesos-executors get launched for each topology is this line:
* https://github.com/mesos/storm/blob/3666d140029a5834c8410b3918d34b2c03343884/src/storm/mesos/MesosNimbus.java#L555

Note that the `ExecutorId` is set to `details.getId()`, and `details` is topology-specific:
* https://github.com/mesos/storm/blob/3666d140029a5834c8410b3918d34b2c03343884/src/storm/mesos/MesosNimbus.java#L424

pdread100:
@erikdw I found my problem. The examples I have been running are the Storm examples and they set the number of workers to 3....I will have to fix that.

pdread100:
@erikdw What version of Storm are you running? I'm running 0.9.3. No matter what I do it hands out enough workers to satisfy the first topo then waits for it finish before assigning slots to the next topo. It never starts a second supervisor, just says there are no more slots available. Even though I have the # ports cranked up, # workers 1, 4 cpus and 4G of mem, even turned off one of the slaves....no joy.

erikdw:
@pdread100 : I'm running 0.9.3 too.  Which sample topologies are you running?  In general storm topologies are meant to run forever, I'm surprised they "finish" for you.  We can email outside of this code review if you want to get it working for you, my address is just my github handle at gmail.

erikdw:
FYI, I've filed a ticket against the Apache Storm project to add support for multiple container-specific logviewers: [STORM-1342](https://issues.apache.org/jira/browse/STORM-1342)

Issue #40, opened 2015-05-19
correct how the storm-mesos tgz file is named 


pdread100:
Currently creates apache-storm...tgz instead of storm-mesos...tgz and when deployed it fails.

erikdw:
This is already fixed:  https://github.com/mesos/storm/commit/5215c5f162bb0c39540fffdb2b7c13839b54d787

PR #42, opened 2015-05-26
Renamed apache-storm-{RELEASE} directory to storm-mesos-{RELEASE} so …

pdread100:
…supervisor could find it, and same with docker image [issue 40] (https://github.com/mesos/storm/issues/40)

This enables the Supervisor to find the correct directory and the dockerImage build-release target as well.


tnachen:
LGTM

PR #45, opened 2015-05-28
Added code checkstyle

pdread100:
Using

mvn validate 

will cause the src to be style checked based on the config/checkstyle.xml. I purposefully turned it off from the normal build-release.sh since it will fail most of the code base. A number of projects seem to tie this to a git pre-commit action instead of the build action.

If this is not something that is found to be useful then I'll close this PR.

### DIFF
--- a/config/checkstyle.xml
+++ b/config/checkstyle.xml
@@ -1,0 +1,321 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE module PUBLIC
+        "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
+        "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
+
+<!-- This is a checkstyle configuration file. For descriptions of
+what the following rules do, please see the checkstyle configuration
+page at http://checkstyle.sourceforge.net/config.html -->
+
+<module name="Checker">
+
+
+    <module name="FileTabCharacter">
+        <!-- Checks that there are no tab characters in the file.
+        -->
+    </module>
+
+    <!--todo:(kgs) add back in-->
+    <!--<module name="NewlineAtEndOfFile"/>-->
+
+    <module name="RegexpSingleline">
+        <!-- Checks that FIXME is not used in comments.  TODO is preferred.
+        -->
+        <property name="format" value="((//.*)|(\*.*))FIXME"/>
+        <property name="message" value='TODO is preferred to FIXME.  e.g. "TODO (johndoe): Refactor when v2 is released."'/>
+    </module>
+
+    <module name="RegexpSingleline">
+        <!-- Checks that TODOs are named.  (Actually, just that they are followed
+             by an open paren.)
+        -->
+        <property name="format" value="((//.*)|(\*.*))TODO [^(]"/>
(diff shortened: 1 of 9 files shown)


pdread100:
The mvn install does not include the -Dcheckstyle.skip=-true, why would it, so I will hard wire a "skip" in the pom.xml and those who want to try it out will have to remove it.

tnachen:
I think if you want to introduce a default style (which I think is great), we'll want to leave it on as default.
Looks like all the errors is valid style rules validations, can you fix them in the same PR? Otherwise I don't see why we add a code style but don't fix the errors.

pdread100:
The intent was to get feedback first. Let folks test drive the style and
say yes or no. However I have no problem with fixing the errors and making
a PR if this style is ok.

config/checkstyle.xml
```diff
@@ -0,0 +1,321 @@
...
+
+        <module name="WhitespaceAfter">
+            <!-- Checks that commas, semicolons and typecasts are followed by
+                 whitespace.
+            -->
+            <property name="tokens" value="COMMA, SEMI, TYPECAST"/>
+        </module>
+
+        <module name="NoWhitespaceAfter">
+            <!-- Checks that there is no whitespace after various unary operators.
```
tnachen: Does this means it only valid to do:

true?1:2

As there is no whitespace allowed?

config/checkstyle.xml
```diff
@@ -0,0 +1,321 @@
...
+        "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
+
+<!-- This is a checkstyle configuration file. For descriptions of
+what the following rules do, please see the checkstyle configuration
+page at http://checkstyle.sourceforge.net/config.html -->
+
+<module name="Checker">
+
+
+    <module name="FileTabCharacter">
```
tnachen: +1

config/checkstyle.xml
```diff
@@ -0,0 +1,321 @@
...
+        NAMING CHECKS
+
+        -->
+
+        <!-- Item 38 - Adhere to generally accepted naming conventions -->
+
+        <module name="PackageName">
+            <!-- Validates identifiers for package names against the
+              supplied expression. -->
+            <!-- Here the default checkstyle rule restricts package name parts to
```
tnachen: I think this is too restrictive. We should remove this

tnachen:
@erikdw @brndnmtthws What you guys think about this code style? IMO we should have one, just the question what styles we should put in. Personally modulo my comments, it seems ok to pull it in and modify it later when we don't like some of them.

brndnmtthws:
It would be nice to standardize this with other projects. Maybe have a look at the Hadoop one:

https://github.com/mesos/hadoop/blob/master/checkstyle.xml

pdread100:
I pulled this style from the mesos/myriad and made some minor mods to the
variable naming since it would not allow underscores at all. A lot of them
put in tabbing which gets pretty ugly.

config/checkstyle.xml
pdread100: Correct, true?1:2 would have to be

true ? 1 : 2

tnachen:
I think this looks ok for now, we can change the style later if we like to

PR #49, opened 2015-06-01
Added dockerized supervisor

pdread100:
This pull contains code to run a supervisor that has been dockerized. 

Moved the docker files into a docker directory
Updated the README
Added a docker image variable to the yaml
Added a new target to create a supervisor docker image to the build scipt
Tested on a two node vagrant cluster


bin/build-release.sh
```diff
@@ -5,6 +5,7 @@ function _rm {
   rm -rf "$@" 2>/dev/null || true
 }
 RELEASE=`grep -1 -A 0 -B 0 '<version>' pom.xml | head -n 1 | awk '{print $1}' | sed -e 's/.*<version>//' | sed -e 's/<\/version>.*//'`
+SUPER_DOCKERTAG=mesos/storm_supervisor:${RELEASE}.ubuntu1404
```
tnachen: I don't see this image in the mesos docker repo?

bin/build-release.sh
```diff
@@ -67,24 +71,37 @@ function package {(
 
   local tarName="storm-mesos-${RELEASE}.tgz"
   cd _release
-  tar czf ${tarName} apache-storm-${RELEASE}
+  # When supervisor starts up it looks for storm-mesos not apache-storm.
+  mv apache-storm-${RELEASE} storm-mesos-${RELEASE}
```
tnachen: Doesn't this conflict with your other PR? That's already merged, so you need to rebase

src/storm/mesos/MesosNimbus.java
```diff
@@ -615,6 +606,57 @@ public void assignSlots(Topologies topologies, Map<String, Collection<WorkerSlot
       }
     }
   }
+  
+  private ExecutorInfo createExecutorInfo(Optional<ContainerInfo.Builder> containerInfoBuilder,
+          TopologyDetails details, String executorDataStr, double executorCpu, double executorMem, 
+          String cpuRole, String memRole, String configUri) {
+              
+        ExecutorInfo.Builder infoBuilder =  ExecutorInfo.newBuilder()
```
tnachen: remove extra space between = and ExecutorInfo

src/storm/mesos/MesosNimbus.java
```diff
@@ -615,6 +606,57 @@ public void assignSlots(Topologies topologies, Map<String, Collection<WorkerSlot
...
+                .addResources(Resource.newBuilder()
+                    .setName("mem")
+                    .setType(Type.SCALAR)
+                    .setScalar(Scalar.newBuilder().setValue(executorMem))
+                    .setRole(memRole));
+                 
+        // If there is a builder then there was a docker image available.
+        if (containerInfoBuilder.isPresent()) {
+            infoBuilder.setCommand(CommandInfo.newBuilder().setShell(false).setUser("vagrant"));
+            infoBuilder.setContainer(containerInfoBuilder.get());
```
tnachen: Why are we hardcoding vagrant user here? We won't be able to merge this with hardcoding a user name here.
And I don't even think vagrant is the default user we should use?

tnachen:
Can you please explain what is a super docker image vs a docker image here? And why you want to do both? Can't we just have one image but pass in different commands for supervisor or something else?

src/storm/mesos/MesosNimbus.java
pdread100: The vagrant was a test on my part and not suppose to be there. When I did a
compare I missed it. I will remove it

bin/build-release.sh
pdread100: Yes it will. I had to have it in there to make everything work....I figured
I would have to rebase anyway.

bin/build-release.sh
pdread100: This just creates it in the local docker repo. I was not sure about pushing
an image to any major repo until it this PR was blessed.

pdread100:
The intent was to create a supervisor image that would be fired up by mesos
when required. I felt it was easier to define the image in this manner
since it was to function strictly as a supervisor. There is nothing
preventing some one from creating other docker images that function however
they want.

tnachen:
And what's the other docker image besides supervisor image?

pdread100:
I don't understand the question? The existing dockerfile would allow in
theory a nimbus image and since it does not specify a command that would
mean, if I understand docker well enough, that the command be specified at
launch, presumably in the command info. The way I did it was much simpler.

I'm don't really see why you are so concerned with the approach since it
addresses the issue of mesos launched docker supervisor very cleanly.

tnachen:
@pdread100 I'm not concerned, it's just not as obvious looking at it because of the naming, that it seems like there are two docker images one that is a parent of another (super image vs image).
We should call it supervisorDockerImage just to be clear, superDockerImage is just too general.
Fix this and we can merge

pdread100:
Cool. Thanks. You are correct its a supervisor image not a super() image.

pdread100:
@tnachen The super has been changed to supervisor. Thanks for your help!

README.markdown
```diff
@@ -63,6 +63,12 @@ bin/build-release.sh downloadStormRelease
 
   Builds a Docker image based on the current packaged storm. Run ./bin/build-release.sh before running this subcommand.
 
+* `superDockerImage`
```
tnachen: I'll fix this myself.

tnachen:
Thanks for the PR!

tnachen:
Actually I can't merge this, there is conflicts now, can you rebase? And fix the comment I have that still refers to superDockerImage

tnachen:
@pdread100

pdread100:
@tnachen I think its ready now. I will contact the mesos folks and see about getting the image added to their registry and create one of my own. I have some docker experts here so I should be able to get this done fairly quickly. Any additional advice would be welcome. Thanks.

src/storm/mesos/MesosNimbus.java
```diff
@@ -617,6 +608,56 @@ public void assignSlots(Topologies topologies, Map<String, Collection<WorkerSlot
...
   }
+  
+  private ExecutorInfo createExecutorInfo(Optional<ContainerInfo.Builder> containerInfoBuilder,
+          TopologyDetails details, String executorDataStr, double executorCpu, double executorMem, 
+          String cpuRole, String memRole, String configUri) {
+              
+        ExecutorInfo.Builder infoBuilder = ExecutorInfo.newBuilder()
+                .setExecutorId(ExecutorID.newBuilder().setValue(details.getId()))
+                .setData(ByteString.copyFromUtf8(executorDataStr))
+                .addResources(Resource.newBuilder()
```
erikdw: Ewww, awesome that you're already pulling this out into a separate function.  I'll make a separate PR to allow having the cpus and mem be 0.  i.e., you cannot add 0 as a resource since the mesos-master will reject it.  We are running with 0 CPUs for the executor because:
* if you dedicate a whole CPU then it's wasteful, the supervisor doesn't really do much
* if you dedicate a franctional CPU then the mesos UI and logs will get totally wonky since there's not great handling for rounding of fractional CPUs (using a resource of 0.1 leads to having available resources be like:  132.999125142456).  At least in mesos-0.20.1 which we are stuck on for now.
* CPUs are the precious resource in our system, we cannot afford to spend hundreds of CPUs on supervisors that do almost nothing.

src/storm/mesos/MesosNimbus.java
```diff
@@ -547,29 +547,20 @@ public void assignSlots(Topologies topologies, Map<String, Collection<WorkerSlot
 
               String executorDataStr = JSONValue.toJSONString(executorData);
               LOG.info("Launching task with Mesos Executor data: <" + executorDataStr + ">");
+
+              // Create a docker builder if a docker image was defined in the yaml
+              Optional<ContainerInfo.Builder> containerInfoBuilder = createDockerInfoBuilder(getDockerImage());
```
erikdw: Why have this in `assignSlots()`?   i.e., if you just call it from within `createExecutorInfo()` then you can avoid this extra parameter, and also slightly shrink the crazy-length of the `assignSlots()` method.

storm.yaml
```diff
@@ -42,7 +46,7 @@ mesos.framework.name: "Storm"
 #
 #mesos.framework.principal: "storm"
 
-# The "secret" phrase cannot be followed by a NL
+# The "secret" phrase cannot be followed by a NewLine
```
erikdw: yay! clarity FTW.  But can you please just use newline instead of NewLine?

src/storm/mesos/MesosNimbus.java
pdread100: Good point. Also I wish some one would refactor assignSlots.....a method this long needs attention.

storm.yaml
pdread100: Sure.

src/storm/mesos/MesosNimbus.java
pdread100: I will leave that in your capable hands since you seem to understand the problem and know what to change. Thanks.

brndnmtthws:
This should no longer be necessary with #65 merged. Feel-free to reopen if you think we still need this.

erikdw:
@brndnmtthws : thanks!  Yeah, your change obviates the need for this one.

PR #53, opened 2015-08-13
Remove supervisor dead code

salimane:
When running
```shell
storm-mesos supervisor
```
I get :
```shell
sh: /opt/storm/bin: is a directory
sh: /opt/storm/bin: is a directory
sh: /opt/storm/bin: is a directory
sh: /opt/storm/bin: is a directory
sh: /opt/storm/bin: is a directory
sh: /opt/storm/bin: is a directory
sh: /opt/storm/bin: is a directory
sh: /opt/storm/bin: is a directory
sh: /opt/storm/bin: is a directory
sh: /opt/storm/bin: is a directory

```
This pr fixes this.

### DIFF
--- a/bin/storm-mesos
+++ b/bin/storm-mesos
@@ -10,10 +10,7 @@ def nimbus():
   os.chdir(STORM_PATH + "/..")
   os.system(STORM_CMD + " nimbus storm.mesos.MesosNimbus")
   
-def supervisor():
-  os.system(STORM_PATH + " supervisor storm.mesos.MesosSupervisor")
-
-COMMANDS = {"nimbus": nimbus, "supervisor": supervisor}
+COMMANDS = {"nimbus": nimbus}
 
 
 def main():


erikdw:
@salimane : if I'm understanding you correctly, you are running "mesos-storm supervisor" yourself?  That's not how mesos-storm is intended to work -- the mesos-slave runs the supervisor.

salimane:
@erikdw, yes "the mesos slave runs the supervisor" but which command is being used to run the supervisor? 

with storm, the command is `bin/storm supervisor`. When you're running storm on top of mesos, what is the command ? I guess it was `bin/storm-mesos supervisor` . If not, why do we need the supervisor function in `bin/storm-mesos`?

how is storm-mesos is intended to work?

Thanks

erikdw:
@salimane : I am not 100% sure of the answer to "how is storm-mesos intended to work".  At my company we build it ourselves in a bit of a hacked-up manner.  I'm trying to understand the stuff that @pdread100 has been doing well enough to support it.

Can you please be a bit more explicit about how you are testing & what build of this project you are using?  I mean, it seems like you aren't *just* using the vagrant setup that @pdread100 put in, because just following the docs/vagrant.md instructions (mostly) works for me without having to make a change like the one you proposed.

salimane:
@erikdw [Launching task with Mesos Executor data](https://github.com/mesos/storm/blob/3666d140029a5834c8410b3918d34b2c03343884/src/storm/mesos/MesosNimbus.java#L560-L561) is interesting. 

I'm not using the included vagrant. I'm doing the setup using a puppet storm module mirroring a real production setup. For a normal storm setup without mesos, you need to run the nimbus ``bin/storm nimbus`` and a supervisor ``bin/storm supervisor`` prior to deploying any topology.

It seems running storm on mesos, the nimbus command become ``bin/storm-mesos nimbus`` and is run before deploying any topology. But then according to the code, there is no need to run any supervisor as a new one would be started while deploying a new topology.

am i right ?

Thanks

erikdw:
@salimane : " But then according to the code, there is no need to run any supervisor as a new one would be started while deploying a new topology."  --> exactly!  With storm-on-mesos there is one Supervisor per topology per host.  The Supervisor *is* the Mesos Executor, and it's launched by the MesosNimbus.  And the storm Worker processes are the Mesos Tasks that run within the container created for the MesosSupervisor.  i.e., on a given host there will be N topologies running in N containers.  1 container per topology per host.  Potentially multiple workers of a given topology within each container.  Containers only ever have a single topologies worker processes, they don't get intermixed.  Hopefully I've stated that exhaustively enough to prevent any confusion.

You can take a gander at some slides I created for a talk I gave at a conference earlier this year for some pictorial views of the differences between regular storm and storm-on-mesos:
* http://gotocon.com/dl/goto-chicago-2015/slides/BrianMcCalister_and_ErikWeathers_EventStreamsAtGrouponWithStormMesosAndGriddle.pdf

You can also watch me talk through the slides here (just found that this was uploaded!):
* https://www.youtube.com/watch?v=TZeHD-z2W6E&t=8m45s

salimane:
@erikdw  thanks for the explanation. But then in the ``bin/storm-mesos`` python script, is the function ``supervisor()`` dead code ?
Thanks

erikdw:
@salimane : the `supervisor()` function in `bin/storm-mesos` appears to have never been used; here's the initial commit of that script (@brndmatthews ported the code from another repo, so it appears he wrote everything, but much of it was written by Nathan Marz, the original Storm author):  
* https://github.com/mesos/storm/blob/b007d54b1a7fad5044932798c1000149d14afd14/bin/storm-mesos#L13-L14

Notably, the invocation of the supervisor wasn't using the `bin/storm-mesos` script then, and isn't doing so now:
https://github.com/mesos/storm/blob/b007d54b1a7fad5044932798c1000149d14afd14/src/storm/mesos/MesosNimbus.java#L400

So I think it's been dead this whole time...

salimane:
@erikdw I've updated the pr to remove the misleading supervisor dead code.
Thanks

brndnmtthws:
Thanks!

erikdw:
thanks for the change @salimane & thanks for merging @brndnmtthws !

PR #59, opened 2015-09-09
improve Mesos task name and set a Mesos executor name; makes it clear…

erikdw:
… in Mesos UI that the tasks are from the storm framework, and also adds the topology-ID into the task name for better searching & sorting within Mesos UI

### DIFF
--- a/src/main/storm/mesos/MesosCommon.java
+++ b/src/main/storm/mesos/MesosCommon.java
@@ -32,8 +32,10 @@ public class MesosCommon {
   public static final String EXECUTOR_MEM_CONF = "topology.mesos.executor.mem.mb";
   public static final String SUICIDE_CONF = "mesos.supervisor.suicide.inactive.timeout.secs";
   public static final String AUTO_START_LOGVIEWER_CONF = "supervisor.autostart.logviewer";
+  // Should we prefix the Worker Task ID with a configurable string (as well as the topology name)?
   public static final String WORKER_NAME_PREFIX = "topology.mesos.worker.prefix";
   public static final String WORKER_NAME_PREFIX_DELIMITER = "topology.mesos.worker.prefix.delimiter";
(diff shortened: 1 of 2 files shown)


erikdw:
I'm planning to change "storm ..." to or "storm-worker ..." in the task name.  And the executor will be changed from "storm ..." to "storm-supervisor ...".

However, I'm wondering if it would be better to do some variation on one of these options for the naming:
~~~
storm-worker-TOPOLOGYID-NODEID:31008
storm-worker_TOPOLOGYID_NODEID:31008
storm-worker__TOPOLOGYID__NODEID:31008
storm-worker::TOPOLOGYID::NODEID:31008
storm-worker TOPOLOGYID NODEID:31008
~~~
NOTE: 
* TOPOLOGYID can have dashes or underscores in it:
* NODEID seems to be the hostname, and that cannot have underscores in it, but will very often have dashes.

I'd love to see some example task name styles that similar frameworks have chosen, if anyone has pointers to some.

erikdw:
@JessicaLHartog: sounds like you like `storm-worker::TOPOLOGYID::NODEID:31008`.   @dsKarthick, do you agree?  Anyone else have an opinion?

dsKarthick:
@erikdw Doesn't look the latest changes. Was expecting ``` storm-worker | TOPOLOGYID | NODEID:31008 ``` format

erikdw:
@dsKarthick : you're right, thanks for reminding me.

erikdw:
I've tested the latest commit (a3d9c04) in my vagrant setup.  Here's what the tasks look like if you change `topology.mesos.component.name.delimiter` to `^^`:

#### delimiter: "^^"

screenshot 2015-12-01 01 29 39

#### As opposed to the default delimiter: " | "


screenshot 2015-12-01 01 38 21

erikdw:
@dsKarthick: please :eyes:

dsKarthick:
:shipit:

PR #63, opened 2015-11-11
[WIP] Scheduler improvements.

brndnmtthws:
- Added additional configuration parameters for controlling scheduler
   behaviour:
   - mesos.offer.filter.seconds
   - mesos.offer.expiry.multiplier
   - mesos.prefer.reserved.resources
 - Implemented combining of reserved & unreserved resources
 - Improved Docker support, added support for running Storm inside
   containers. This also introduces the mesos.container.docker.image
   config param
 - Added unit tests
 - Improved port handling, especially with regard to logviewer
 - Code cleanup/style fixes
 - Implemented filters & reviveOffers()

NOTE: this isn't quite ready to merge yet.

PR #65, opened 2015-11-11
[WIP] Storm improvements

brndnmtthws:
- Added additional configuration parameters for controlling scheduler
   behaviour:
   - mesos.offer.filter.seconds
   - mesos.offer.expiry.multiplier
   - mesos.prefer.reserved.resources
 - Implemented combining of reserved & unreserved resources
 - Improved Docker support, added support for running Storm inside
   containers. This also introduces the mesos.container.docker.image
   config param
 - Added unit tests
 - Improved port handling, especially with regard to logviewer
 - Code cleanup/style fixes
 - Implemented filters & reviveOffers()

Also from @maverick2202:

 - Upgrade Storm
 - Add worker name prefix

NOTE: this isn't quite ready to merge yet.

Let's merge this instead of #62 and #63.

cc @sargun @erikdw


src/main/storm/mesos/LaunchTask.java
```diff
@@ -0,0 +1,46 @@
...
+  public Protos.TaskInfo getTask() {
+    return task;
+  }
+
+  public Protos.Offer getOffer() {
+    return offer;
+  }
+
+  private final Protos.TaskInfo task;
+  private final Protos.Offer offer;
```
sargun: These should be at the top of the class.

src/main/storm/mesos/MesosCommon.java
```diff
@@ -31,14 +32,44 @@
...
+  public static final double DEFAULT_EXECUTOR_MEM_MB = 500;
   public static final int DEFAULT_SUICIDE_TIMEOUT_SECS = 120;
 
   public static final String SUPERVISOR_ID = "supervisorid";
   public static final String ASSIGNMENT_ID = "assignmentid";
-
+  public static final String DEFAULT_DELIMITER = "_";
+  
+  public static String hostFromAssignmentId(String assignmentId, String delimiter) {
+    int last = assignmentId.lastIndexOf(delimiter);
```
sargun: declare final

src/main/storm/mesos/MesosCommon.java
```diff
@@ -31,14 +32,44 @@
...
+    String host = assignmentId.substring(last + delimiter.length());
+    LOG.debug("AssignMentId: " + assignmentId + " Host: " + host);
+    return host;
+  }
+  
+  public static String getWorkerPrefix(Map conf, TopologyDetails info) {
+    conf = getFullTopologyConfig(conf, info);
```
sargun: I have a preference for single assignment. This is just a nicety.

src/main/storm/mesos/MesosNimbus.java
```diff
@@ -67,13 +60,19 @@
...
+  public static final String CONF_MESOS_LOCAL_FILE_SERVER_DIR = "mesos.local.file.server.dir";
   public static final String CONF_MESOS_FRAMEWORK_NAME = "mesos.framework.name";
+  public static final String CONF_MESOS_PREFER_RESERVED_RESOURCES = "mesos.prefer.reserved.resources";
+  public static final String CONF_MESOS_CONTAINER_DOCKER_IMAGE = "mesos.container.docker.image";
 
-  public static final Logger LOG = Logger.getLogger(MesosNimbus.class);
+  private static final Logger LOG = Logger.getLogger(MesosNimbus.class);
 
-  private static final String FRAMEWORK_ID = "FRAMEWORK_ID";
+  public static final String FRAMEWORK_ID = "FRAMEWORK_ID";
```
sargun: Define public statics at the top of file

src/main/storm/mesos/MesosNimbus.java
```diff
@@ -84,16 +83,17 @@
   Optional<Integer> _localFileServerPort;
   private RotatingMap<OfferID, Offer> _offers;
   private LocalFileServer _httpServer;
-  private java.net.URI _configUrl;
   private Map<TaskID, Offer> _usedOffers;
   private ScheduledExecutorService timerScheduler =
       Executors.newScheduledThreadPool(1);
+  private boolean _preferReservedResources = true;
+  private String _container = null;
```
sargun: Any reason not to use an Optional here?

src/main/storm/mesos/MesosNimbus.java
```diff
@@ -114,94 +123,221 @@ public String getHostName(Map<String, SupervisorDetails> map, String nodeId) {
...
+
+  public void doRegistration(final SchedulerDriver driver, Protos.FrameworkID id) {
+    _driver = driver;
+    try {
+      _state.put(FRAMEWORK_ID, id.getValue());
+    } catch (IOException e) {
+      LOG.error("Halting process...", e);
+      Runtime.getRuntime().halt(1);
+    }
+    Number filterSeconds = (Number) _conf.get(CONF_MESOS_OFFER_FILTER_SECONDS);
```
sargun: Use Optional here Optional<Number>.fromNullable(...).or(120)

src/main/storm/mesos/MesosNimbus.java
```diff
@@ -383,461 +631,263 @@ public void assignSlots(Topologies topologies, Map<String, Collection<WorkerSlot
...
+    List<WorkerSlot> workerSlots = slotList.get(id);
+    boolean usingExistingOffer = false;
+    boolean subtractedExecutorResources = false;
+
+    for (WorkerSlot slot : workerSlots) {
+      TaskID taskId = TaskID.newBuilder()
+          .setValue(MesosCommon.taskId(slot.getNodeId(), slot.getPort()))
+          .build();
+
+      if (id == null || offer == null && _usedOffers.containsKey(taskId)) {
```
sargun: Please put parens around this to be explicit about comparison order

src/main/storm/mesos/MesosNimbus.java
```diff
@@ -209,10 +345,6 @@ private OfferResources getResources(Offer offer, double executorCpu, double exec
         }
       }
     }
-
-    LOG.debug("Offer: " + offerToString(offer));
-    LOG.debug("Extracted resources: " + resources.toString());
-    return resources;
   }
 
   private List<WorkerSlot> toSlots(Offer offer, double cpu, double mem, boolean supervisorExists) {
```
sargun: Do you combine offers if you received multiple offers from the same agent?

src/main/storm/mesos/MesosNimbus.java
```diff
@@ -114,94 +123,221 @@ public String getHostName(Map<String, SupervisorDetails> map, String nodeId) {
...
+  public void shutdown() throws Exception {
+    _httpServer.shutdown();
+  }
+
+  public void resourceOffers(SchedulerDriver driver, List<Protos.Offer> offers) {
+    synchronized (_offersLock) {
+      LOG.debug("resourceOffers: Currently have " + _offers.size() + " offers buffered" +
+          (_offers.size() > 0 ? (":" + offerMapToString(_offers)) : ""));
+      for (Protos.Offer offer : offers) {
+        if (_offers != null && isHostAccepted(offer.getHostname())) {
```
sargun: If _offers was null, it would have thrown a null pointer exception at line 221

sargun:
Please add findbugs to the pom:
```
    <reporting>
      <plugins>
        <plugin>
          <groupId>org.codehaus.mojo</groupId>
          <artifactId>findbugs-maven-plugin</artifactId>
          <version>3.0.3</version>
          <configuration>
            <effort>Max</effort>
            <threshold>Default</threshold>
            <xmlOutput>true</xmlOutput>
          </configuration>
        </plugin>
      </plugins>
    </reporting>
```

src/main/storm/mesos/MesosNimbus.java
brndnmtthws: :)

src/main/storm/mesos/MesosNimbus.java
brndnmtthws: Not across offer cycles, no.

brndnmtthws:
Thanks for the review @sargun. Updated the PR as per your comments.

src/main/storm/mesos/MesosNimbus.java
sargun: Within a given "cycle" -- where you're holding on to offers, are you combining them? A given slave can split up its resources.

bin/storm-mesos
```diff
@@ -9,7 +9,7 @@ STORM_CMD = STORM_PATH + "/storm"
 def nimbus():
```
sargun: Please PEP8 this file:
```
3c075477e55e:bin sdhillon$ pep8 storm-mesos 
storm-mesos:9:1: E302 expected 2 blank lines, found 1
storm-mesos:10:3: E111 indentation is not a multiple of four
storm-mesos:11:3: E111 indentation is not a multiple of four
```

src/main/storm/mesos/MesosNimbus.java
brndnmtthws: It does combine them, but only if the OfferID matches.

src/main/storm/mesos/MesosCommon.java
```diff
@@ -31,13 +32,46 @@
...
   public static final int DEFAULT_SUICIDE_TIMEOUT_SECS = 120;
 
   public static final String SUPERVISOR_ID = "supervisorid";
   public static final String ASSIGNMENT_ID = "assignmentid";
+  public static final String DEFAULT_DELIMITER = "_";
+  
+  public static String hostFromAssignmentId(String assignmentId, String delimiter) {
+    final int last = assignmentId.lastIndexOf(delimiter);
+    String host = assignmentId.substring(last + delimiter.length());
+    LOG.debug("AssignMentId: " + assignmentId + " Host: " + host);
```
sargun: AssignMentId? Why is the M capitalized?

brndnmtthws: Not sure. I'll fix it though.

sargun:
All the places you're doing:
if (X == null)
   X = $default_value

Why not use Optionals across the board?

src/main/storm/mesos/MesosNimbus.java
```diff
@@ -114,94 +120,229 @@ public String getHostName(Map<String, SupervisorDetails> map, String nodeId) {
...
+    }
+  }
+
+  public void offerRescinded(OfferID id) {
+    synchronized (_offersLock) {
+      _offers.remove(id);
+    }
+  }
+
+  public void taskLost(final TaskID taskId) {
```
sargun: If you're not doing reconcilation, how will this callback ever get triggered on disconnection?

brndnmtthws:
Think I got all the ` == null`'s.

src/main/storm/mesos/MesosNimbus.java
brndnmtthws: Hmm, I'm not sure why you're asking. This could be called any time there's a status update with `TASK_LOST`.

sargun: Yeah - but you're not doing task reconciliation. So, if you're dependent on task statuses working at all, things wont work well.

brndnmtthws: Fair enough. Storm does its own out-of-band reconciliation-like thing (using ZK), so I'm not too worried about it.

sargun:
I was unable to test fault recovery, but it looks pretty good. Please reorder / squash commits.

README.markdown
```diff
@@ -88,28 +84,29 @@ Storm/Mesos provides resource isolation between topologies. So you don't need to
...
 
 3. `storm.zookeeper.servers`: The location of the Zookeeper servers to be used by the Storm master.
 
-4. `nimbus.host`: The hostname of where you run Nimbus.
-
 ## Optional configuration
 
 * `mesos.supervisor.suicide.inactive.timeout.secs`: Seconds to wait before supervisor to suicides if supervisor has no task to run. Defaults to "120".
-* `mesos.master.failover.timeout.secs`: Framework failover timeout in second. Defaults to "3600".
+* `mesos.master.failover.timeout.secs`: Framework failover timeout in second. Defaults to "24*7*3600".
```
dsKarthick: @brndnmtthws Whats the motivation for setting such a long timeout?

README.markdown
```diff
@@ -88,28 +84,29 @@ Storm/Mesos provides resource isolation between topologies. So you don't need to
...
 * `mesos.allowed.hosts`: Allowed hosts to run topology, which takes hostname list as a white list.
 * `mesos.disallowed.hosts`: Disallowed hosts to run topology, which takes hostname list as a back list.
 * `mesos.framework.role`: Framework role to use. Defaults to "*".
 * `mesos.framework.checkpoint`: Enabled framework checkpoint or not. Defaults to false.
 * `mesos.offer.lru.cache.size`: LRU cache size. Defaults to "1000".
+* `mesos.offer.filter.seconds`: Number of seconds to filter unused Mesos offers. Defaults to "120".
```
dsKarthick: What does filtering unsued offers mean? Just by reading this I am not able to grok the purpose of this config.

README.markdown
brndnmtthws: To prevent accidental framework removal. It's just a default, so you can always specify something else :)

README.markdown
brndnmtthws: I looked for some docs on this...but couldn't find anything. The tl;dr is: the allocator will not send this same offer to the framework (Storm) for the period specified, unless the framework calls `reviveOffers()`. It will call `reviveOffers()` when the offers are actually needed.

This makes the framework play nicer with other frameworks in a multi-tenant cluster.

brndnmtthws: I'll update the docs though to make this more clear.

README.markdown
```diff
@@ -117,7 +114,54 @@ Storm/Mesos provides resource isolation between topologies. So you don't need to
 * `topology.mesos.worker.mem.mb`: Memory (in MiB) per worker. Defaults to "1000".
   * `worker.childopts`: Use this for JVM opts.  You should have about 25% memory overhead for each task.  For
   example, with `-Xmx1000m`, you should set `topology.mesos.worker.mem.mb: 1250`
-* `topology.mesos.executor.cpu`: CPUs per executor. Defaults to "1".
-* `topology.mesos.executor.mem.mb`: Memory (in MiB) per executor. Defaults to "1000".
+* `topology.mesos.executor.cpu`: CPUs per executor. Defaults to "0.1".
```
dsKarthick: Do we really need to change this to 0.1? This would break the contract for all the existing users. By users, I mean those who are running topologies on mesos/storm cluster.

storm.yaml
```diff
@@ -4,27 +4,36 @@
...
 # -----------------------------------------------------------
 
 
-### If this is not false then you need to add about 128m of mem usage to executor to account for the logviewer
-#supervisor.autostart.logviewer: false
-topology.mesos.executor.mem.mb: 1128  ## default is 1000
+# Worker resources
+topology.mesos.worker.cpu: 1.0
+### Worker heap with 20% overhead
+topology.mesos.executor.mem.mb: 1250
```
dsKarthick: did you intend to do ```topology.mesos.worker.mem.mb: 1250```

brndnmtthws: Looks like that should be `1200`. Good catch.

erikdw:
@brndnmtthws : thanks for doing the move separately.  Can you please also apply style changes separately?  Clubbing together functional changes with cosmetic changes obscures what's being done.  Also, you just have a ton of different things being done at once, I suppose it's ok, since this isn't like the most rigorous of projects.

brndnmtthws:
@erikdw I'm not super happy about all the changes in one PR, but I have to pick my battles :)

storm.yaml
```diff
@@ -4,27 +4,36 @@
...
+# If you enable the logviewer, you'll need to add memory overhead to the
+# executor for the logviewer.
 logviewer.childopts: "-Xmx128m"
 logviewer.cleanup.age.mins: 10080
 logviewer.appender.name: "A1"
-
-
-# You should not need to change anything below this line
-#--------------------------------------------------------
+supervisor.autostart.logviewer: true
```
erikdw: We should disable logviewer by default -- it's totally broken for any system with more than 1 topology, and the solution is unclear:
* https://github.com/mesos/storm/pull/38#issuecomment-111259423
* https://github.com/mesos/storm/issues/6#issuecomment-112218318
* https://github.com/mesos/storm/pull/38#issuecomment-112215933

storm.yaml
```diff
@@ -34,7 +43,7 @@ storm.local.dir: "storm-local"
 # role must be one of the mesos-master's roles defined in the --roles flag
 #
 mesos.framework.role: "*"
-mesos.framework.checkpoint: false
+mesos.framework.checkpoint: true
```
erikdw: Cool, I'm looking forward to trying checkpointing -- I didn't even know about it until attending MesosCon.

storm.yaml
brndnmtthws: I believe this should be fixed now, actually. Previously the port was hardcoded (and fixed), hence the conflict. It'll now bind to separate ports for each supervisor.

storm.yaml
erikdw: That wasn't the big issue, it was `executor` instead of `worker`. ;-)

storm.yaml
brndnmtthws: Although--it just occurred to me that I'm not sure how _Nimbus_ handles the ports. I made the assumption that the supervisor reports back to Nimbus where it's running, but that may be invalid.

erikdw: Ah... I haven't read everything you've changed here.  But I have now skimmed through the logviewer parts -- you seem to be grabbing another port for per-executor/supervisor logviewers and launching them appropriately.  But as you just guessed, there is no interaction between the nimbus nor storm-UI and the logviewer about what port the logviewer has on a particular host for a particular topology.   From grepping the storm master branch, we can see that it is just 1 global config knob for the logviewers' port:

```
storm-core/src/clj/backtype/storm/daemon/logviewer.clj:          https-port (int (or (conf LOGVIEWER-HTTPS-PORT) 0))
storm-core/src/clj/backtype/storm/daemon/logviewer.clj:      (storm-run-jetty {:port (int (conf LOGVIEWER-PORT))
storm-core/src/clj/backtype/storm/ui/core.clj:  (if (and secure? (*STORM-CONF* LOGVIEWER-HTTPS-PORT))
storm-core/src/clj/backtype/storm/ui/core.clj:      (*STORM-CONF* LOGVIEWER-HTTPS-PORT)
storm-core/src/clj/backtype/storm/ui/core.clj:      (*STORM-CONF* LOGVIEWER-PORT)
storm-core/src/clj/backtype/storm/ui/core.clj:  (logviewer-link host (event-logs-filename topology-id port) secure?))
storm-core/src/clj/backtype/storm/ui/core.clj:  (url-format "http://%s:%s/daemonlog?file=nimbus.log" host (*STORM-CONF* LOGVIEWER-PORT) port))
storm-core/src/clj/backtype/storm/ui/core.clj:              (*STORM-CONF* LOGVIEWER-PORT)
storm-core/src/jvm/backtype/storm/Config.java:    public static final String LOGVIEWER_PORT = "logviewer.port";
storm-core/src/jvm/backtype/storm/Config.java:     * Storm Logviewer HTTPS port
storm-core/src/jvm/backtype/storm/Config.java:    public static final String LOGVIEWER_HTTPS_PORT = "logviewer.https.port";
```

brndnmtthws: That's rather unfortunate. The other possibility is to use a reserved port for the logviewer, and only launch the logviewer when the port is available. You could use roles/reservations to guarantee the port is available to Storm across the cluster, but that seems like a pretty lame solution to me. It doesn't doesn't solve the multiple-supervisors problem.

brndnmtthws: I spent some time this morning going through the Storm source to see how much effort it'd be to add the logviewer port to the supervisor heartbeats.

Suffice to say, all the parentheses made it a little difficult to figure out what's what. I might take another stab at it later.

Since it'd require patching Storm, I think I'll make it so the logviewer will launch only if the port is available, and otherwise not. This should let you run multiple supervisors, but not from the same Storm cluster.

erikdw: Yeah, the storm clojure is especially hard due to the large number of storm-specific macros and the un-functional nature of it (tons of state usage and inter-dependencies and long functions).

To be clear, I don't consider this to be supervisor-specific so much as Mesos Executor-specific;  i.e., the executor containers have their separate sandboxes for the log files, and those need to be exposed separately.  Unless you do as my company does and send your logs out of the container into the native host's filesystem.  Then you can have a single logviewer.

Can you please clarify: "This should let you run multiple supervisors, but not from the same Storm cluster."?   We will *definitely* still be running multiple topologies on each host from a single storm cluster overlay.  And I honestly don't care about the logviewer...

brndnmtthws: What I meant was that you could not have 2 supervisors from the same nimbus, per IP (or host machine)--since they'd both try to bind a logviewer to the same port.

erikdw: To be clear about our terminology, the "supervisor" (the actual storm supervisor daemon) doesn't know anything about the logviewer.  It's just another daemon in the storm universe that runs on the host where the worker processes spit out their logs.   Now, the MesosSupervisor (which is the Mesos Executor here) *can* choose to launch the logviewer as we are apparently doing here.  But that's not an inherent storm thing by any means.

So long as this MesosSupervisor-launched logviewer is disable-able (as it is right now) I suppose it's an ok compromise, since as you've noted, Storm isn't making it easy on us to have multiple logviewers per host.

brndnmtthws: Yes, that's correct :)

src/main/storm/mesos/MesosSupervisor.java
```diff
@@ -112,7 +108,7 @@ public String getSupervisorId() {
 
   @Override
   public String getAssignmentId() {
-    return _assignmentId;
+    return MesosCommon.hostFromAssignmentId(_assignmentId, MesosCommon.getWorkerPrefixDelimiter(_conf));
```
erikdw: I still do not understand *why* @maverick2202 wants to change this assignmentId stuff.  Please explain *why* we are adding this extra logic.

Dockerfile
```diff
@@ -1,22 +1,22 @@
...
 ENV JAVA_HOME /usr/lib/jvm/java-7-openjdk-amd64
 
 ENV MESOS_NATIVE_JAVA_LIBRARY /usr/lib/libmesos.so
 
-# copy local checkout into /opt/storm
-ADD . /opt/storm
+ADD . /work
 
+WORKDIR /work
+RUN ./bin/build-release.sh main
```
erikdw: I'm not a docker expert, but if I'm reading this right then, the Docker image will *build* the release every time it is run.  Why is this being done?

I would really appreciate explicit enumeration of *all* behavior changes and their raison d'etre in this extensive PR.

e.g., why was ENTRYPOINT removed below?

brndnmtthws: `ENTRYPOINT` was removed because it makes it possible to pass environment variables via shell commands. With `ENTRYPOINT`, you can pass additional arguments, but they're passed as literal strings (so passing an env var doesn't work).

brndnmtthws:
Updated the PR to use fixed ports for the logviewer.

Dockerfile
erikdw: Sorry, but I'm still a bit lost, I blame my inexperience with Docker.  My understanding is that a Docker image should (must?) have a "thing" to be run when the image's container launches, where this "thing" is specified via either ENTRYPOINT or CMD.  But I don't see a CMD either.

I guess this implies that when we run the resultant Docker image we now *must* specific the cmd to be executed (somehow).  However, I don't see any docs in this PR about how to do this for this repo (meaning both the docker syntax and the image-resident cmd to be run).  Can you please add those?

(BTW, after overcoming my initial forgetfulness above about the purpose of the Dockerfile, I totally support this pulling of the `build-release.sh` invocation into the Dockerfile, instead of having it be a separate external step.)

brndnmtthws: If, for some reason, I wanted to run a shell inside the container I would execute:

```shell
docker run -i --net=host -t brndnmtthws/storm /bin/bash
```

Where `/bin/bash` above is the command I'm running inside the container (i.e., the entrypoint). It's a little different from specifying an explicit entrypoint (either via the Dockerfile, or with --entrypoint).

The tl;dr is that you don't _need_ to specify an entrypoint or command in the Dockerfile. In fact, I prefer _not_ specifying it. That kind of metadata belongs in the Marathon definition, rather than the container image, IMO.

README.markdown
```diff
@@ -88,36 +84,86 @@ Storm/Mesos provides resource isolation between topologies. So you don't need to
 
 ## Mandatory configuration
 
-1. `mesos.executor.uri`: Once you fill in the configs and repack the distribution, you need to place the distribution somewhere where Mesos executors can find it. Typically this is on HDFS, and this config is the location of where you put the distibution.
+1. `mesos.executor.uri` or `mesos.container.docker.image`: Once you fill in the configs and repack the distribution, you need to place the distribution somewhere where Mesos executors can find it. Typically this is on HDFS, and this config is the location of where you put the distibution. Alternatively, you may use a Docker image in place of the executor URI. Take a look at the Dockerfile in the top-level of this repository for an example of how to use it.
```
JessicaLHartog: This would be easier to understand if it was more like:
* One of the following:
 * `mesos.executor.uri`: Once you fill in the configs and repack the distribution, you need to place the distribution somewhere where Mesos executors can find it. Typically this is on HDFS, and this config is the location of where you put the distibution.
 * `mesos.container.docker.image`: `<description similar to the one above>`

Also provide a description of what happens if you specify both.

brndnmtthws: Good point! I'll update accordingly.

README.markdown
```diff
@@ -88,36 +84,86 @@ Storm/Mesos provides resource isolation between topologies. So you don't need to
...
 * `mesos.offer.lru.cache.size`: LRU cache size. Defaults to "1000".
+* `mesos.offer.filter.seconds`: Number of seconds to filter unused Mesos offers. These offers may be revived by the framework when needed. Defaults to "120".
+* `mesos.offer.expiry.multiplier`: Offer expiry multiplier for `nimbus.monitor.freq.secs`. Defaults to "2.5".
 * `mesos.local.file.server.port`: Port for the local file server to bind to. Defaults to a random port.
 * `mesos.framework.name`: Framework name. Defaults to "Storm!!!".
 * `mesos.framework.principal`: Framework principal to use to register with Mesos
 * `mesos.framework.secret.file`:  Location of file that contains the principal's secret. Secret cannot end with a NL.
-* `supervisor.autostart.logviewer`: Default is true, if not false please add 128M to topology.mesos.executor.mem.mb
+* `mesos.prefer.reserved.resources`: Prefer reserved resources over unreserved (i.e., `"*"` role). Defaults to "true".
+* `supervisor.autostart.logviewer`: Default is true, if not false please add 128M to `topology.mesos.executor.mem.mb`.
```
JessicaLHartog: The default here is true, so it would make more sense for the default of `topology.mesos.executor.mem.mb` to include these 128 MB, and have the message here instead be "Default is true; if false, you may subtract 128 MB form `topology.mesos.executor.mem.mb`".

brndnmtthws: I'll clarify this message.

Dockerfile
erikdw: @brndnmtthws : thanks for always being so prompt with responses!   Yup, that docker cmd syntax looks familiar.

Can we at least be explicit in the documentation about the cmd that needs to be run if you are doing Docker without Marathon.  i.e., I can see from this PR that the cmd is `/opt/storm/bin/storm-mesos` (I suppose with supervisor or nimbus as options for *what* to run), but I don't see any reference to it in the "green portion" of the PR.

brndnmtthws: I'll make a note in the README. Frankly, running it without Marathon isn't much different :) You could just run the same command that's being passed through to Marathon. There's a small wrapper script which handles ports, but that's about it.

brndnmtthws:
Updated as per the feedback.

README.markdown
```diff
@@ -88,36 +84,86 @@ Storm/Mesos provides resource isolation between topologies. So you don't need to
...
-* `topology.mesos.executor.mem.mb`: Memory (in MiB) per executor. Defaults to "1000".
-  * `supervisor.childopts`: Use this for executor (aka supervisor) JVM opts.  You should have about 25% memory
-  overhead for each task.  For example, with `-Xmx500m`, you should set `topology.mesos.executor.mem.mb: 625`
+  * `worker.childopts`: Use this for JVM opts.  You should have about 20% memory overhead for each task.  For
+    example, with `-Xmx1000m`, you should set `topology.mesos.worker.mem.mb: 1200`. By default this is platform
+    dependent.
+* `topology.mesos.executor.cpu`: CPUs per executor. Defaults to "0.1".
+* `topology.mesos.executor.mem.mb`: Memory (in MiB) per executor. Defaults to "500".
+  * `supervisor.childopts`: Use this for executor (aka supervisor) JVM opts.  You should have about 20% memory
+    overhead for each task.  For example, with `-Xmx500m`, you should set `topology.mesos.executor.mem.mb: 625`. By
```
JessicaLHartog: The math here is wrong. 
500*1.20 = 600 != 625.

Out of curiosity, is there any reason in particular that this suggestion decreased from 25% to 20% for both `worker.childopts` and `topology.mesos.executor.mem.mb`?

brndnmtthws: Good catch, fixed that.

I ran some testing and found that the 25% recommendation was too high. It's just a recommendation, so YMMV. Certainly 25% is more conservative than 20%, and it really depends on workload.

brndnmtthws: Let me update the wording in the docs on the overhead. I'll say 20-25%, but leave the defaults as 20%. Sound good?

Dockerfile
erikdw: @brndnmtthws : thanks for the additional docs!

README.markdown
JessicaLHartog: That sounds good. :+1:

src/main/storm/mesos/MesosNimbus.java
```diff
@@ -114,94 +120,228 @@ public String getHostName(Map<String, SupervisorDetails> map, String nodeId) {
...
+
+  public void doRegistration(final SchedulerDriver driver, Protos.FrameworkID id) {
+    _driver = driver;
+    try {
+      _state.put(FRAMEWORK_ID, id.getValue());
+    } catch (IOException e) {
+      LOG.error("Halting process...", e);
+      Runtime.getRuntime().halt(1);
+    }
+    Number filterSeconds = Optional.fromNullable((Number) _conf.get(CONF_MESOS_OFFER_FILTER_SECONDS)).or(120);
```
erikdw: I see you're only using this "filter" config here -- should we also use it for the `launchTasks()` invocation in `computeLaunchList()`?

erikdw: Also, maybe the magic number of 120 be put into a default constant variable in MesosCommon.java, and the value be returned using a wrapper function, to mimic the existing style?

src/main/storm/mesos/MesosNimbus.java
```diff
@@ -114,94 +120,228 @@ public String getHostName(Map<String, SupervisorDetails> map, String nodeId) {
...
+                val.getId(),
+                filters
+            );
+          }
+        }
+    );
+
+    Number lruCacheSize = Optional.fromNullable((Number) _conf.get(CONF_MESOS_OFFER_LRU_CACHE_SIZE)).or(1000);
+    final int intLruCacheSize = lruCacheSize.intValue();
+    _usedOffers = Collections.synchronizedMap(new LinkedHashMap<Protos.TaskID, Protos.Offer>(intLruCacheSize + 1, .75F, true) {
```
erikdw: So, we've realized something about this `_usedOffers` cache when we were reading through the code the other day.  It's effectively pointless if we're not persisting `_usedOffers` and then reading it back upon framework scheduler (MesosNimbus) restart.  Right?

src/main/storm/mesos/MesosNimbus.java
brndnmtthws: I think adding it to `launchTasks()` might not give the behaviour we want. We don't want to start filtering offers until we know there are no more workers to launch.

Maybe rather than piling more code changes into this PR, I can do a cleanup of the config handling in a follow-up one.

src/main/storm/mesos/MesosNimbus.java
brndnmtthws: Yeah, it's best effort, at best. I think it's okay, since it's really just an optimization.

erikdw: fair enough -- I may readdress this logic in the future when I come back to the "avoid hoarding" technique we want to try.

src/main/storm/mesos/MesosNimbus.java
erikdw: Thought provoking response... let me ensure we are on the same page about what filtering means, when it's being applied, and what you are trying to accomplish (I'm sorry about the lack of brevity here, I've spent a bunch of time editing this down, it's a complex subject).

Here is my understanding of the purpose of `Filters.refuse_seconds` (attained after sitting in a zen-like pose and contemplating the deepest secrets of the universe):

* the functionality controls how fast a framework might get a "refused" resource back from the mesos-master, after having chosen not to use it.  And the choice to not use (thus "refuse") a resource is made after having been offered that resource, and the choice is made by either:
 1. a call to `declineOffer()` which declines all resources in the referenced Offer, OR
 2. a call to `launchTasks()` where the launched task(s) don't use *all* of the resources in the chosen Offer(s), thus "refusing" the unused resources.
* even without your changes we are *already* filtering in both cases, since based on [this comment in the mesos.proto file](https://github.com/apache/mesos/blob/0.25.0/include/mesos/mesos.proto#L1166-L1171), the behavior applies to all "unused resources" with a 5 second timeout, whether or not you have chosen to pass Filters explicitly in your call to `declineOffer()` or `launchTasks()`

So my *current* understanding is that we are "filtering" by default even if we are not explicitly choosing to configure the time that we "filter" for.  It's just that the default is for 5 seconds, and you are leaving that alone for `launchTasks()`, and instead setting it to something much larger (120 seconds) for `declineOffer()`.

So now let me reconcile this understanding with your changes and statement about "only filter when we know there are no more workers to launch":

* The default of your new config knob being 120 seconds (way bigger than the 5 second default) implies that when you say "we don't want to start filtering offers until we know there are no more workers to launch", you really mean: "we don't want to filter offers *for a really long time* unless we know there are no more workers to launch".  I conclude that because we *are* filtering for 5 seconds even in the case of `launchTasks()`, so we *are* filtering offers even today whenever we call `declineOffer()` & `launchTasks()`
* Here's my understanding of your desire: you want to back off of receiving any resources back from previously received offers when those offers haven't been used for the TTL of waiting through the RotatingMap.   And *if* we end up needing offers back we will call `reviveOffers()` and then in the next round-trip through the Nimbus scheduling we'll get the resources back.  This seems to be a nice solution to the "greedy hoarding" behavior of the MesosNimbus. Pretty cool.  This feature might obviate the need to directly implement the "only hoard when topologies are pending" idea.

I guess this is all a very long-winded way of saying "it makes sense to do this".  With a couple caveats:

1. We should be strict in our phrasing when discussing this feature.  It's not that we are *only* filtering for declined offers.  It's that we're only filtering for a "configured potentially longer period of time" for declined offers that have timed out in our RotatingMap.  Not sure how to phrase that nicely.  But it's *not* that we're *only* filtering for decline offers.
2. The mesos.proto docs and/or API docs should be improved a bit to make the purpose of `Filters.refuse_seconds` more clear.  i.e., "unused resources" is insufficiently explicated, as technically the offers being hoarded by MesosNimbus are also "unused", and thus I was getting confused thinking that this "refuse_seconds" functionality was related to rescinding/InverseOffers.  Perhaps if the comment simply said "refused resources" it would be clearer.  Or an explanation that "unused resources" *does* **not** include those offers that are being held by a framework and are as-of-yet unused.

src/main/storm/mesos/MesosNimbus.java
```diff
@@ -114,94 +120,228 @@ public String getHostName(Map<String, SupervisorDetails> map, String nodeId) {
...
+      // This method is called just after a new entry has been added
+      public boolean removeEldestEntry(Map.Entry eldest) {
+        return size() > intLruCacheSize;
+      }
+    });
+
+    Number offerExpired = Optional.fromNullable((Number) _conf.get(Config.NIMBUS_MONITOR_FREQ_SECS)).or(60);
```
erikdw: I realize the default value of 60 seconds was already there, I just want to explicitly call out that [Storm's default value for this config param is 10 seconds](https://github.com/apache/storm/blob/v0.10.0/conf/defaults.yaml#L60)

src/main/storm/mesos/MesosNimbus.java
```diff
@@ -114,94 +120,228 @@ public String getHostName(Map<String, SupervisorDetails> map, String nodeId) {
...
+        try {
+          synchronized (_offersLock) {
+            _offers.rotate();
+          }
+        } catch (Throwable t) {
+          LOG.error("Received fatal error Halting process...", t);
+          Runtime.getRuntime().halt(2);
+        }
+      }
+    }, 0, Math.round(1000 * expiryMultiplier.doubleValue() * offerExpired.intValue()));
```
erikdw: ok, so this slightly changes the previous timing.  It used to be 2000*10 (by default) == 20 seconds.  It's not 1000*2.5*10 == 25 seconds.  I think that's ok, just calling it out.

erikdw:
@brndnmtthws : we're almost done with reviewing this, at least based on the claimed features in the description box on the Conversation page.  Need to still understand these:
* mesos.prefer.reserved.resources
* Implemented combining of reserved & unreserved resources
* assignmentId change that are in this PR (i.e., the change ported from https://github.com/mesos/storm/pull/62)

And also read through the changes in their entirety.

src/main/storm/mesos/MesosNimbus.java
erikdw: Though can you please explicitly comment in the code about the intent of this optimization?  It seemed to be to be used to issue "killTask" calls, but those didn't seem to have much practical use to me based on my understanding and witnessing of what happens in our systems.

src/main/storm/mesos/MesosNimbus.java
brndnmtthws: Yes, that's correct. I guess it can be thought of as 'additional filtering' for when then offers aren't required.

src/main/storm/mesos/MesosNimbus.java
brndnmtthws: It's used for the purpose of reusing existing slots (or offers). In other words, if a worker was already launched with an offer, but we can now repurpose that slot (or offer), it will first kill the task then launch a new one.

To be honest, I'm not even sure if this feature works. If we're going to change that, let's do it as part of a separate PR.

src/main/storm/mesos/MesosNimbus.java
brndnmtthws: Noted. I'll update the default.

brndnmtthws:
@erikdw Okay, thanks for the info.

A note on reserved & unreserved: If you use reservations (the Mesos feature), offers will arrive with both unreserved (i.e., role='*') and reserved (role='storm') resources. These resources can be combined (from the same host). Furthermore, if there are both reserved and unreserved resources, you need to choose whether to prefer reserved or unreserved resources when computing which resources to consume first.

Let me know if you need more clarification.

src/main/storm/mesos/MesosNimbus.java
erikdw: +1, let's readdress later.

config/checkstyle.xml
```diff
@@ -9,313 +9,303 @@ page at http://checkstyle.sourceforge.net/config.html -->
...
+         by an open paren.)
+    -->
+    <property name="format" value="((//.*)|(\*.*))TODO [^(]"/>
+    <property name="message" value='All TODOs should be named.  e.g. "TODO (johndoe): Refactor when v2 is released."'/>
+  </module>
+
+  <module name="RegexpSingleline">
+    <!-- Check for trailing whitespace. -->
+    <!-- \s matches whitespace character, $ matches end of line. -->
+    <property name="format" value="\s+$"/>
```
JessicaLHartog: Can we add `<property name="message" value='Please remove trailing whitespace.'/>` here?

brndnmtthws:
I'm going to go ahead and merge this now.

If there are any further comments/suggestions, I'll address them in a separate PR.

erikdw:
DOOOOOOOOOD!!!

erikdw:
I said we had other comments man.  This is a HUGE change.

brndnmtthws:
Happy to address them quickly. I just don't want this PR to keep growing over time, because we'll never get it merged.

erikdw:
Well... let's all make an effort to keep changes small and modular in the future then.  Then they are reviewable without herculean effort, less likely to break things, etc.

config/checkstyle.xml
```diff
@@ -9,313 +9,304 @@ page at http://checkstyle.sourceforge.net/config.html -->
...
 
-        WHITESPACE CHECKS
+    -->
 
+    <module name="WhitespaceAround">
+      <!-- Checks that various tokens are surrounded by whitespace.
+             This includes most binary operators and keywords followed
+             by regular or curly braces.
         -->
+      <property name="tokens" value="ASSIGN, BAND, BAND_ASSIGN, BOR,
```
JessicaLHartog: The difference between this, and just accepting the default are the following values:
* `DO_WHILE` (the `while` keyword in a do-while)
* `LCURLY` (left curly) 
* `LITERAL_SWITCH` (the `switch` keyword)
* `RCURLY`(right curly)
* `SLIST` (a statement list)
* `LITERAL_ASSERT` (the `assert` keyword)
* `TYPE_EXTENSION_AND` (the `&` symbol when used in a generic upper or lower bounds constrain)

If there isn't a particular reason for excluding these values, this entire block can be simplified to `module name="WhitespaceAround"/>`

config/checkstyle.xml
```diff
@@ -9,313 +9,304 @@ page at http://checkstyle.sourceforge.net/config.html -->
...
+              BOR_ASSIGN, BSR, BSR_ASSIGN, BXOR, BXOR_ASSIGN, COLON, DIV, DIV_ASSIGN,
+              EQUAL, GE, GT, LAND, LE, LITERAL_CATCH, LITERAL_DO, LITERAL_ELSE,
+              LITERAL_FINALLY, LITERAL_FOR, LITERAL_IF, LITERAL_RETURN,
+              LITERAL_SYNCHRONIZED, LITERAL_TRY, LITERAL_WHILE, LOR, LT, MINUS,
+              MINUS_ASSIGN, MOD, MOD_ASSIGN, NOT_EQUAL, PLUS, PLUS_ASSIGN, QUESTION,
+              SL, SL_ASSIGN, SR_ASSIGN, STAR, STAR_ASSIGN"/>
+      <property name="severity" value="error"/>
+    </module>
+
+    <module name="WhitespaceAfter">
```
JessicaLHartog: There is no difference between this and the default, so the block can be simplified to `<module name="WhitespaceAfter"/>

config/checkstyle.xml
```diff
@@ -9,313 +9,304 @@ page at http://checkstyle.sourceforge.net/config.html -->
...
+      <!-- Checks that commas, semicolons and typecasts are followed by
+           whitespace.
+      -->
+      <property name="tokens" value="COMMA, SEMI, TYPECAST"/>
+    </module>
+
+    <module name="NoWhitespaceAfter">
```
JessicaLHartog: The difference between this, and just accepting the default are the following values:
* `ARRAY_INIT` (An array initialization)
* `ARRAY_DECLARATOR`(An array declaration) 
* `INDEX_OP`(The array index operator)

If there isn't a particular reason for excluding these values, this entire block can be simplified to `<module name="NoWhitespaceAfter"/>`

config/checkstyle.xml
```diff
@@ -9,313 +9,304 @@ page at http://checkstyle.sourceforge.net/config.html -->
...
-            <property name="severity" value="error"/>
-        </module>
-
-        <module name="ParenPad">
-            <!-- Checks that there is no whitespace before close parens or after
-                 open parens.
-            -->
-            <property name="severity" value="warning"/>
-        </module>
+    <module name="NoWhitespaceBefore">
```
JessicaLHartog: Again, this is only leaving out `COMMA` and explicitly allowing line breaks. The specification is better by default, and in fact, picks up the horrendous spacing on `src/main/storm/mesos/MesosNimbus.java:407` that the custom settings fail to see.

I suggest replacing the block with `<module name="NoWhitespaceBefore"/>`

config/checkstyle.xml
```diff
@@ -9,313 +9,304 @@ page at http://checkstyle.sourceforge.net/config.html -->
...
+      <!-- Checks that there is no whitespace before various unary operators.
+           Linebreaks are allowed.
+      -->
+      <property name="tokens" value="SEMI, DOT, POST_DEC, POST_INC"/>
+      <property name="allowLineBreaks" value="true"/>
+      <property name="severity" value="error"/>
+    </module>
+
+    <module name="ParenPad">
```
JessicaLHartog: Same as above, I suggest replacing this block with `<module name="ParenPad"/>`

Dockerfile
erikdw: @brndnmtthws : ok, so I'm getting back to testing with the docker stuff in this repo -- it seems your changes have totally busted the `bin/build-release.sh dockerImage` option -- so I'm guessing you didn't notice that option and are just running the docker build manually.  I did it like so:
```
docker build -t mesos/storm:git-`git rev-parse --short HEAD` .
```

Something I don't like about this new style of buliding docker images is that it isn't caching any of the java artifacts being downloaded, so it makes rebuilds slow.  Any ideas for improving that behavior?

brndnmtthws: AFAIK there isn't a good way to cache the artifacts. Furthermore, we're just using the Dockerhub automated builds: https://hub.docker.com/r/mesosphere/storm/

Here's a PR to cut down the image size: https://github.com/mesos/storm/pull/74

erikdw: Ah, so when you say "using the Dockerhub automated builds" you're referring to the Github "service/hook" for Docker to kick off those builds, as linked here:  https://github.com/mesos/storm/settings/hooks/6434119

brndnmtthws: Yes, that's correct.

erikdw:
I'm not sure yet what the cause is, but when I'm testing the post-#65 version of this project, it is unable to launch multiple worker tasks in the same executor at the same time.  I'm suspicious of the assignmentId changes, but haven't figured out the cause yet.  As an example I have a small topology that needs 3 workers, all 3 of which get assigned to the same host (I only have 1 mesos-slave host) per the MesosNimbus logs, but only 1 comes up every 2 minutes.  i.e., first port 31000 comes up, but for the other 2 there are no logs in the supervisor, then a bit over 2 minutes later it's 31001 that comes up, then a bit over 2 minutes after that it's 31002 that comes up.   Will update once I figure out more about what's happening, but figured I'd mention it as soon as I validated that it's happening.

src/main/storm/mesos/MesosNimbus.java
```diff
@@ -383,461 +635,269 @@ public void assignSlots(Topologies topologies, Map<String, Collection<WorkerSlot
...
-            @Override
-            public void expire(OfferID key, Offer val) {
-              driver.declineOffer(val.getId());
-            }
+        String executorDataStr = JSONValue.toJSONString(executorData);
+        ExecutorInfo.Builder executorInfoBuilder = ExecutorInfo.newBuilder();
+        executorInfoBuilder
+            .setExecutorId(ExecutorID.newBuilder().setValue(details.getId()))
+            .setData(ByteString.copyFromUtf8(executorDataStr));
+        if (!subtractedExecutorResources) {
```
erikdw: This is the logic causing [the problem I described here](https://github.com/mesos/storm/pull/65#issuecomment-161906189).  I'm sending a PR in a second.

src/main/storm/mesos/MesosNimbus.java
```diff
@@ -114,94 +120,228 @@ public String getHostName(Map<String, SupervisorDetails> map, String nodeId) {
...
     for (Resource r : offer.getResourcesList()) {
-      if (r.getName().equals("cpus") && r.getScalar().getValue() > offerCpu) {
-        offerCpu = r.getScalar().getValue();
-        resources.cpuSlots = (int) Math.floor((offerCpu - executorCpu) / cpu);
-      } else if (r.getName().equals("mem") && r.getScalar().getValue() > offerMem) {
-        offerMem = r.getScalar().getValue();
-        resources.memSlots = (int) Math.floor((offerMem - executorMem) / mem);
+      if (r.hasReservation()) {
+        // skip resources with reservations
+        continue;
```
dsKarthick: @brndnmtthws Why are we skipping reserved resources?

brndnmtthws: With the way dynamic reservations are implemented in Mesos, you may receive reserved offers for other frameworks. Since the Storm framework doesn't implement dynamic reservations, we just decline all of them.

PR #67, opened 2015-11-18
Clean up checkstyle config.

brndnmtthws:
to address @JessicaLHartog's comments from #65.

### DIFF
--- a/config/checkstyle.xml
+++ b/config/checkstyle.xml
@@ -235,58 +235,14 @@ page at http://checkstyle.sourceforge.net/config.html -->
     </module>
 
 
-    <!--
+    <!-- WHITESPACE CHECKS -->
 
-    WHITESPACE CHECKS
-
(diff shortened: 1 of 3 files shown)


JessicaLHartog:
@brndnmtthws Could youalso do the same thing for `GenericWhitespace` at the end of `checkstyle.xml`? Looks like it's [not really configurable](http://checkstyle.sourceforge.net/config_whitespace.html#GenericWhitespace), and the proposed change just makes error messages go from something like this:
```
<file path>:<line number>:<column number>: GenericWhitespace '<' is followed by whitespace.
```
to something like this:
```
<file path>:<line number>:<column number>: '<' is followed by whitespace.
```

brndnmtthws:
Sure! Done.

JessicaLHartog:
Awesome LGTM. :+1:

Issue #76, opened 2015-12-04
storm-mesos versioning


erikdw:
@brndnmtthws @tnachen @DarinJ @dskarthick @jessicalhartog

I'm opening this issue to create a place for us to brainstorm and decide on a new versioning scheme for this project.

##### Current version scheme
The version is currently just directly set to the version of storm being "used".  Where used means:
* included as build dependency of the java bits in this project
* bundling of the storm binary bits into this project's resultant package

This unfortunately doesn't allow us to vary the version of this project as we make changes to the code that are independent of the storm version.

##### Simplistic storm-independent versioning
The first idea is to simply set the version at say `0.1.0` to start, and then we increment the version as we make changes to the code.  However, this limits us to having only ever 1 version of storm and mesos within a given "version" of the code.

##### Supporting multiple storm versions
Per [this discussion in #68](https://github.com/mesos/storm/issues/68#issuecomment-161337875), we are planning to support multiple versions of storm from a given version of the code in this project.  This implies that we should be somehow tagging / naming / versioning the resultant package with the storm version used.

##### Supporting multiple mesos versions
We could also potentially build against multiple mesos versions, in case there are backwards incompatible changes.

##### Docker image versioning
I'm admittedly ignorant about Docker image versioning, but when I look at the [builds on Dockerhub](https://hub.docker.com/r/mesosphere/storm/builds/), I cannot even figure out what commit SHA the image was built from.  I feel like we can do better with the versioning there, ideally taking advantage of the conclusion of this discussion.

##### Proposal

Use maven classifiers and [follow the maven artifact description scheme](https://maven.apache.org/plugins-archives/maven-deploy-plugin-2.6/examples/deploying-with-classifiers.html) of suffixing the artifact description with the classifier string. Whoa that's a mouthful, let me give an example:

Basically, an example classifier could be:  `storm0.9.6`.  Then the artifact & resultant package could be called:
* maven artifact: `storm-mesos-0.1.0-storm0.9.6.jar`
* built package: `storm-mesos-0.1.0-storm0.9.6.tar.gz`

Then if we decided to start tweaking the mesos version as well, we could make this even longer:
* built package: `storm-mesos-0.1.0-storm0.9.6-mesos0.25.0.tar.gz`

DarinJ:
[...]

Issue #68, opened 2015-11-21
Support for storm-0.10.0


DarinJ:
Storm 0.10.0 changed a few things that break storm on mesos.  In particular:
- LocalState.get now returns a TBase object (thrift related) instead of String.
- bin/storm is now a bash script instead of python
- snakeyaml is no longer in the classpath
- commons-lang upgraded to common-lang3 which effects imports

I spent a little time correcting these, and have a really rough working version [here](https://github.com/darinj/storm/tree/0.10.0).  I'll probably clean it up next week as I need it, if there's interest I'll submit a PR.  This does break storm 0.9.X though so it may be good to write a shim or place on a separate branch/tag.

erikdw:
@DarinJ : thanks for the research and prototype.  I haven't looked at the issues with storm-0.10.0 yet, but definitely wanna keep storm-0.9.x compatibility for now.   Not sure how hard a "shim" would be to write, but I suppose we could use maven profiles to separate out inclusion of some code / dependencies.

salimane:
@erikdw @DarinJ we can use tag to deal with the compatibility issue

erikdw:
@salimane : can you please elaborate on how using git tags would help deal with the issue?

i.e., we cannot just create a "storm-0.9.6" tag and freeze the code in its current form for when people run storm-0.9.6.  We have more changes to be made to the framework.  

A separate git branch could work, but requires double the work.   Ideally we would have a maven profile and/or some sort of bridge class.

DarinJ:
@erikdw @salimane I think a separate branch is way too much work and it almost always ends up breaking something.  I've looked at the effort to do profiles for this and it's pretty trivial, just gotta find time to do it.  I hope to have something next week (maybe the week after).

To give you an idea of my plan:
- Have two storm-local-state-0.XX maven subprojects. They will contain a class implementing an interface wrapping `LocalState` to expose `get`, `put` and `remove` and a class which exposes the proper command line for the executor/supervisor. The storm-mesos project will use the correct subproject based off profile.  This solves 1 & 2. 
- Add a dependency for snakeyaml in the 0.10 profile, solving 3.
- Replace `StringUtils.isNotEmpty` from commons-lang with `!Strings.isNullorEmpty` from guava and remove the commons.lang import. This fixes 4. 

Open to suggestions.  Will also need to mod the build.sh script haven't thought that though yet.

erikdw:
[...]

PR #77, opened 2015-12-04
WIP - Adds support for storm-0.10.0

DarinJ:
Per Issue #68 Adds support for storm-0.10.0

This creates a multi-module maven build and a v9-v10 shim.  Profiles are used to choose the correct shim at compile time.  Modified build-release.sh to use STORM_RELEASE and MESOS_RELEASE environment variables to influence build behavior.

./build-release.sh builds 0.9.6 and mesos 0.24.1
STORM_RELEASE=0.X.X builds for storm 0.X.X and mesos 0.24.1
STORM_RELEASE=0.X.X MESOS_RELEASE=0.Y.Y builds for storm 0.X.X and mesos 0.Y.Y

Updated versioning of artifacts per issue #76.
Updated .gitignore for _docker, *.zip, *.tgz, to avoid polluting git with build artifacts.

Tested storm-0.9.6 and storm-0.10.0 on live clusters.  No issues to date.


erikdw:
@DarinJ : I see you've added the classifier stuff and adjusted the framework tarball with the storm and mesos suffixes -- awesome!   It *looks* like it's in good shape, should I test this out now, or are you intending on iterating on it a bit more first?

DarinJ:
@erikdw yeah I'm kind conservative on the @'s.  I also sneakily edited my original post to make it even harder to figure out :).

I just got around to testing 0.9.6 and 0.10.0 and they're both doing as expected.  Please test, especially if you can test the docker image, I'm not setup to do that presently.  If you're happy enough to merge, I can squash the commits and rebase.

pom.xml
```diff
@@ -13,67 +14,64 @@
...
   </properties>
 
+  <profiles>
+    <profile>
+      <id>storm9</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
+      </activation>
+      <properties>
+        <storm.version>0.9.3</storm.version>
```
choang: baseline to 0.9.6

pom.xml
```diff
@@ -94,86 +92,91 @@
...
+    </dependencies>
+  </dependencyManagement>
+
   <reporting>
     <plugins>
       <plugin>
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>findbugs-maven-plugin</artifactId>
-        <version>3.0.3</version>
+        <version>3.0.1</version>
```
choang: why revving down on this?

storm-shim/src/main/java/storm/mesos/shims/ICommandLineShim.java
```diff
@@ -0,0 +1,12 @@
+package storm.mesos.shims;
+
+/**
+ * Returns the command line, shimmed and 0.9.X and 0.10.X have slightly a different command
+ */
+public interface ICommandLineShim {
+
+  public String getCommandLine();
+
+  public String getDockerCommandLine();
```
choang: the interface without ``getDockerCommandLine()` makes more sense to me.  I was expecting to see the following concrete classes:
- `StandardCommandLineShim implements ICommandLineShim`
- `DockerCommandLineShim implements ICommandLineShim`

In the `MesosNimbus` code, you could have:
```java
ICommandLineShim  commandLineShim = Factory.makeCommandLineShim(commandType);
...
  .setCommand(commandLineShim.getCommandLine())
...
```

I think you can avoid having the two separate invocations of `setCommand` with this approach.
```

File: pom.xml (hunk 1 above)
Comment (DarinJ) [inline pom.xml]: Yep,  will fix all those before final cut.

File: pom.xml (hunk 2 above)
Comment (DarinJ) [inline pom.xml]: Will revert, it was a temporary fix to an ide issue.

File: storm-shim/src/main/java/storm/mesos/shims/ICommandLineShim.java (hunk 3 above)
Comment (DarinJ) [inline storm-shim/src/main/java/storm/mesos/shims/ICommandLineShim.java]: The Factory class can certainly live in the storm-mesos module so there's really no more maven work.  Aesthetically it may be more pleasing.  The one negative is there's slightly more code for each new shim (storm 0.11.0?) with @choang's approach, while the Mesos-Nimbus code stays constant with either approach.  Are there strong opinions here?

Comment (choang) [inline storm-shim/src/main/java/storm/mesos/shims/ICommandLineShim.java]: I think going with the factory approach is just a start for refactoring `computeResourcesForSlot()`.  That method does way more than just compute the resources.

For the near term, the `MesosNimbus` code could go from:
```java
        if (_container.isPresent()) {
          // An ugly workaround for a bug in DCOS
          Map<String, String> env = System.getenv();
          String javaLibPath = env.get("MESOS_NATIVE_JAVA_LIBRARY");

          executorInfoBuilder
              .setCommand(CommandInfo.newBuilder()
                  .addUris(URI.newBuilder().setValue(configUri))
                  .setValue(
                      "export MESOS_NATIVE_JAVA_LIBRARY=" + javaLibPath +
                          " && /bin/cp $MESOS_SANDBOX/storm.yaml conf && /usr/bin/python bin/storm " +
                          "supervisor storm.mesos.MesosSupervisor -c storm.log.dir=$MESOS_SANDBOX/logs" + extraConfig))
              .setContainer(
                  ContainerInfo.newBuilder()
                      .setType(ContainerInfo.Type.DOCKER)
                      .setDocker(
                          ContainerInfo.DockerInfo.newBuilder()
                              .setImage(_container.get())
                              .setNetwork(ContainerInfo.DockerInfo.Network.HOST)
                              .setForcePullImage(true)
                              .build()
                      )
                      .build()
            );
        } else {
          executorInfoBuilder
              .setCommand(CommandInfo.newBuilder()
                  .addUris(URI.newBuilder().setValue((String) _conf.get(CONF_EXECUTOR_URI)))
                  .addUris(URI.newBuilder().setValue(configUri))
                  .setValue("cp storm.yaml storm-mesos*/conf && cd storm-mesos* && python bin/storm " +
                      "supervisor storm.mesos.MesosSupervisor" + extraConfig));
        }
```
to
```java
    ICommandLineShim  commandLineShim = Factory.makeCommandLineShim(commandType);
          executorInfoBuilder
              .setCommand(CommandInfo.newBuilder()
                  .addUris(URI.newBuilder().setValue((String) _conf.get(CONF_EXECUTOR_URI)))
                  .addUris(URI.newBuilder().setValue(configUri))
                  .setValue(commandLineShim.getCommandLine()));
        if (_container.isPresent()) {
          executorInfoBuilder
              .setContainer(
                  ContainerInfo.newBuilder()
                      .setType(ContainerInfo.Type.DOCKER)
                      .setDocker(
                          ContainerInfo.DockerInfo.newBuilder()
                              .setImage(_container.get())
                              .setNetwork(ContainerInfo.DockerInfo.Network.HOST)
                              .setForcePullImage(true)
                              .build()
                      )
                      .build()
            );
        }
```

I don't have a strong opinion, but I do think that future refactoring will make it so the more OO design will be used.

erikdw:
@DarinJ : I just merged PR #80 that conflicts a bit with this PR (#77).  Can you please rebase / merge your changes on top of that?  Thanks!

pom.xml
```diff
@@ -3,9 +3,10 @@
   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.apache.mesos</groupId>
-  <artifactId>storm</artifactId>
-  <version>0.9.6</version>
-  <name>storm</name>
+  <artifactId>storm-parent</artifactId>
+  <version>0.1.0</version>
```
erikdw: I'm thinking we'll want to use a `-SNAPSHOT` suffix in the version, allowing us to directly use the `maven release` plugin.  Or we could fake its use by manually rev'ing the version.  Thoughts?

DarinJ: @erikdw I'm pro-releases vs pull from master and hope for the best.  I'm not sure how well the release plugin and build-release.sh (it greps the pom for version) will play together, likely we'll have to fake it.

I kicked around the idea of getting maven to do the storm download and tarball surgery. It's doable and would get the exact result you want.  I don't presently have time to do it, but could help review/advise.  Might be able to get me interest in January/February.  Should definitely be separate PR anyway.

P.S: will rebase and incorporate @choang this evening/tomorrow.  Do you want snapshot added to the version for now and will just mod it for releases?

erikdw: 1. Yeah, we'll have to finagle the system to get the `build-release.sh` working with the release/assembly plugins of maven, won't work right out of the box.
2. I also don't have time, my zone is about to be called in the airport, going on vacation. ;-)
3. Yeah, let's just put -SNAPSHOT in for the moment.

DarinJ:
@erikdw @choang OK, rebased, applied requested changes and retested.  I think we're good.  Noticed one issue after merge in the new version of build-release.sh. In the dockerImage `build -t xxx` failed had to move to `build -t xxx .`.

storm-shim/src/main/java/storm/mesos/shims/ICommandLineShim.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+package storm.mesos.shims;
+
+/**
+ * Returns the command line, shimmed and 0.9.X and 0.10.X have slightly a different command
+ */
+public interface ICommandLineShim {
+
+  public String getCommandLine();
```
choang: in the interest of reducing the lines of code, could you make `extraConfig` a parameter here?  You can make `extraConfig` optional by using an abstract class that implements an overloaded method, however, I don't see the optional being necessary with this use case.

choang: Just so I am more clear, the goal is to allow you to remove the constructors and `extraConfig` as instance variables from each of the concrete implementations.

DarinJ: I agree it could be improved but I'm not sure it's worth the effort here.  It's a shim, it lives in a very small area of the code and isn't going to be touched again until storm 0.11.0 (if it needs to be).

DarinJ:
@brndnmtthws erikdw was looking at this but it on vacation, can you look at this.  I tested it quite a bit, seems to work fine.  After #85 I noticed it's getting more difficult to use `LocalState` on the supervisor as it was previously without a lot of effort, I tried to maintain the current feel of LocalState, but it might be easier to use a `HashSet` for keeping track of the ports.

DarinJ:
@erikdw hope you had a nice vacation.  When you get a chance can you look at this.

erikdw:
@DarinJ : yep, will do, thanks for your work on it so far, I'll build it with 0.9.6 and verify it's working.

BTW, the "latest" mesos is now 0.26.0, I had some issues with my docker image and vagrant VM host having differing mesos versions (i.e., `mesos-slave` was setting `MESOS_NATIVE_JAVA_LIBRARY` to `/usr/lib/libmesos-0.26.0.so`, instead of just `/usr/lib/libmesos.so`, and since my docker image had `/usr/lib/libmesos-0.25.0.so` it failed).  I'm guessing working around such inconsistencies is the reason for [@brndnmtthws's workaround here.](https://github.com/mesos/storm/blob/f7ae20832771f8991c8873b8da4c899474127710/src/main/storm/mesos/MesosNimbus.java#L798-L806), unless there was a bug fixed in `mesos-0.26.0` regarding passing of env variables / setting of that particular one for docker.

erikdw:
@DarinJ : I've tested with storm 0.9.6 and mesos 0.26.0.  FYI, I had to make 1 other change for using mesos 0.26.0:

```
-          credentialBuilder.setSecret(ByteString.readFrom(new FileInputStream(secretFilename)));
+          credentialBuilder.setSecret(ByteString.readFrom(new FileInputStream(secretFilename)).toString());
```

Also, I had to make a change to the various vagrant scripts to allow this new naming scheme to work:
`${RELEASE}` -> `${RELEASE}*`

Can you please change make the vagrant script change along with your other changes in here?

Also, give me a bit of time to read through the code a bit more please.

erikdw:
(Just leaving a breadcrumb here)

I wondered what happened with `LocalState` in storm-0.10.0 that required this shim to be created.  Here's the explanation:

storm-0.9.6 implementation of `LocalState` was based on Java serialization and the `snapshot` is a map from `Object` to `Object`:
* https://github.com/apache/storm/blob/v0.9.6/storm-core/src/jvm/backtype/storm/utils/LocalState.java

Whereas in storm-0.10.0, `LocalState` has been changed to use Thrift for serialization and the `snapshot` is a map from `String` to `TBase`:
* https://github.com/apache/storm/blob/v0.10.0/storm-core/src/jvm/backtype/storm/utils/LocalState.java

This was done to allow for rolling upgrades that avoid problems with the stored state being inconsistent between storm versions:
* https://issues.apache.org/jira/browse/STORM-765
* https://issues.apache.org/jira/browse/STORM-634

Here are the changes made to storm that implemented this new behavior:
* https://github.com/apache/storm/commit/b1dc422bd502ad031d027ba0c026c3e666abe42f#diff-437638c20d4ef66fa4e866a8316c3919

storm-shim-10x/src/main/java/storm/mesos/shims/LocalStateShim.java
```diff
@@ -0,0 +1,72 @@
...
+  LocalState _state;
+
+  public LocalStateShim(String localDir) throws IOException {
+    _state = new LocalState(localDir);
+  }
+
+  public String get(String key) {
+    TBase tBase = _state.get(key);
+    if (tBase != null) {
+      JavaObject jo = ((JavaObject) tBase);
```
erikdw: how did you figure out this logic for translating the `TBase` to a `String`?

storm-shim-10x/src/main/java/storm/mesos/shims/LocalStateShim.java
```diff
@@ -0,0 +1,72 @@
...
+      }
+    }
+    return null;
+  }
+
+  public void put(String key, String value) {
+    ArrayList<JavaObjectArg> tmpList = new ArrayList<JavaObjectArg>();
+    tmpList.add(JavaObjectArg.string_arg(value));
+    _state.put(key, new JavaObject("java.lang.String", tmpList));
+
```
erikdw: extraneous newline

storm-shim-9x/src/main/java/storm/mesos/shims/DockerCommandLineShim.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class DockerCommandLineShim implements ICommandLineShim {
+  String extraConfig;
+
+  public DockerCommandLineShim(String extraConfig) {
+    this.extraConfig = extraConfig;
+  }
+
+  public String getCommandLine() {
+    // An ugly workaround for a bug in DCOS
```
erikdw: I'm not convinced this stuff is needed if you're using consistent mesos versions on the host and in the docker container.  But we can leave it for now.

storm-shim-10x/src/main/java/storm/mesos/shims/LocalStateShim.java
DarinJ: Unfortunately, Storm didn't update any documentation about LocalState (still says references types <Object,Object>, so dug through the code.  There's probably a better way.  After doing this, I considered using a `Map<String, String>` in the `Supervisor.java` and using Zookeeper (or the mesos state apis) to keep track of the frameworkId in `Nimbus.java`.  This would eliminate the need for `LocalState` altogether.  This worked though so didn't want to bother until discussing I didn't know how you'd feel about those changes.

storm-shim-9x/src/main/java/storm/mesos/shims/DockerCommandLineShim.java
DarinJ: I think the idea was you'd be running the nimbus and the supervisor both in docker containers using the same image.  In that circumstance this avoids having to hardcode the path.  I don't know if I like it, as I might want to run nimbus w/o docker and but run the supervisors in docker but have different paths to libmesos.so.  I doubt I'd get much sympathy for that setup though.

storm-shim-10x/src/main/java/storm/mesos/shims/LocalStateShim.java
erikdw: Do you have any references to the Mesos State APIs' docs?  I've found some references to some frameworks that supposedly use them (e.g., in Aurora), but I've had a really hard time finding some simple API.  i.e., I expect a trivial API with add/get/delete into the Mesos State managed by the mesos-master.  But I think I might misunderstand what they are providing with the "State API" and that it is simply providing an interface layer on top of a ton of junk you have to set up yourself.

erikdw: Regarding using ZooKeeper for state instead -- yeah, for Supervisor seems overkill (if they die we just bring up another one, they don't need to recover state across restarts), but for Nimbus could be useful for supporting HA Nimbus, although Storm 0.11 will have HA Nimbus in-built so that might not be needed.

DarinJ:
@erikdw bit of a delayed response due to holidays.  Glad to know it worked for you, I'm taking a look a the vagrant scripts now seems straight forward and can do that.

Looked at the changes for mesos 0.26.0, looks like the `CredentialsBuilder.setSecret` method now has an argument of type `String` vs `ByteString`, and they created a second method `CredentialsBuilder.setSecretBytes` which takes a `ByteString`.  We could do yet another shim here but for something this simple and the fact it's only accessed once reflection seems like the better choice (though I'm sure I'll eventually regret it).  I've already coded that and am in the process of testing.

Not sure I'll get a chance to do any new changes to LocalState for a bit and while it could be better it works.

erikdw:
@DarinJ : ah, so you're gonna use reflection to figure out whether we need to use the `String` vs. `ByteString` in the code, thus preserving support for pre-0.26.0 and post-0.26.0 mesos.  Cool.

DarinJ:
@erikdw Think I got all of your requested changes in.  There's a few things that could be better, but I think it's best to address them separately (They're already mentioned as issues elsewhere).  They are:
- Not using LocalState, instead use Zookeeper on nimbus and a ConcurrentMap on supervisors
- Use maven to build tar ball and docker images, use maven-release plugin
- Investigate best way to pass MESOS_NATIVE_LIB to to supervisors when using Docker

drewrobb:
This is working for me in my test environment with docker and 0.26. :+1:

erikdw:
Thanks for the changes @DarinJ!  Minor nits mentioned above and then we should be good.  BTW, I would also love it if you happened to know more about the "mesos state storage" APIs that [I mentioned above.](https://github.com/mesos/storm/pull/77#discussion_r48629130)  We can take that offline though.

DarinJ:
@erikdw address squashed commits and rebased.  Still need to test the docker build since merging (not overly worried but want to be thorough), so give me the weekend to sign off.  

I did notice during the merge on line 101 of the build-release.sh this line got changed in the last PR:

```
+  cmd="docker build -t karthick/mesos-storm:git-`git rev-parse --short HEAD` ."
```

Not sure if that was intentional, pinging @dsKarthick also.  Let me know if it should go back to mesos I can revert that change.  Might be wise to make it $DOCKER_USERNAME passed in like STORM_RELEASE and MESOS_RELEASE, so they can push the built image to dockerhub?  Might be wise to do DOCKER_REGISTRY too in case they want to push to a local registry for speed.

erikdw:
Oh, that was surely a mistake by @dsKarthick & me -- we don't want his own dockerhub account being used, so yes, please revert that.

Let's handle generalizing the docker config in a separate PR to unblock this one.

I'll hold off on your a-ok after the weekend.

Thanks a lot for all the work on this!

DarinJ:
@erikdw built the docker image and all went well, should be good changes prior to rebasing have been working well for me so far.  Left the change to the docker tag as a separate commit for book keeping.

erikdw:
thanks @DarinJ !!   Let me sanity check tomorrow, assume it's all good though so should be merging this in tomorrow.

DarinJ:
@erikdw gentle reminder. If there's issues I'd like to know while it's still in my head :).

erikdw:
@DarinJ : thanks for the reminder.  I just didn't prioritize completing the testing well.  I've now tested with storm-0.10.0 & mesos-0.26.0, looks good to me!  I'm going to merge your change now, thank you very much for your hard work on making this framework a bit version-agnostic and implementing some of our ideas for versioning and build-ability.

erikdw:
FYI, I tested with storm-0.9.6 and mesos-0.25.0 and it worked as well.  Woot!

PR #78, opened 2015-12-04
allow simultaneous launching of 2+ worker processes per host

erikdw:
One of the logic changes in PR #65 broke the ability to simultaneously launch
more than 1 worker process for a given topology.  The cause of the breakage
was intentionally avoiding inclusion of the executor's resources into the
ExecutorInfo structure associated with the mesos tasks (storm workers).
This is problematic because the mesos-master rejects tasks whose ExecutorInfo
isn't identical to other tasks under the same executor.  Notably, since the
executor is already running for subsequent tasks, the resources that are
added to these subsequent tasks' ExecutorInfo aren't actually used, so there
is no advantage in attempting to avoid their inclusion.

FFR, this is the commit with that change:
* af8c49beac04b530c33c1401c829caaa8e368a35

After this fix I was able to instantly launch 3 workers for a topology
on the same mesos-slave host.

### DIFF
--- a/src/main/storm/mesos/MesosNimbus.java
+++ b/src/main/storm/mesos/MesosNimbus.java
@@ -789,15 +789,11 @@ public class MesosNimbus implements INimbus {
         executorInfoBuilder
             .setName(executorName)
             .setExecutorId(ExecutorID.newBuilder().setValue(details.getId()))
-            .setData(ByteString.copyFromUtf8(executorDataStr));
-        if (!subtractedExecutorResources) {
-          subtractedExecutorResources = true;
-          executorInfoBuilder
-              .addAllResources(executorCpuResources)
-              .addAllResources(executorMemResources);
-          if (executorPortsResources != null) {
-            executorInfoBuilder.addAllResources(executorPortsResources);
-          }
+            .setData(ByteString.copyFromUtf8(executorDataStr))
+            .addAllResources(executorCpuResources)
+            .addAllResources(executorMemResources);
+        if (executorPortsResources != null) {
+          executorInfoBuilder.addAllResources(executorPortsResources);
         }
         if (_container.isPresent()) {
           // An ugly workaround for a bug in DCOS


erikdw:
@brndnmtthws & @dskarthick & @jessicalhartog: please :eyes: when you have :watch:

brndnmtthws:
Ah, makes sense.

:shipit:

brndnmtthws:
Makes sense. :shipit:

erikdw:
Thanks @brndnmtthws.  FFR, [this is the comment where I explored this problem and its cause](https://github.com/mesos/storm/pull/65#issuecomment-161906189).

src/main/storm/mesos/MesosNimbus.java
```diff
@@ -789,15 +789,11 @@ protected void computeResourcesForSlot(final RotatingMap<OfferID, Offer> offers,
         executorInfoBuilder
             .setName(executorName)
             .setExecutorId(ExecutorID.newBuilder().setValue(details.getId()))
-            .setData(ByteString.copyFromUtf8(executorDataStr));
-        if (!subtractedExecutorResources) {
-          subtractedExecutorResources = true;
```
JessicaLHartog: I think we need to add this assignment back in. With this change it no longer gets set to true in the rest of the code, and if `subtractedExecutorResources` means the same as it did before PR #65, this will mean that we're over-provisioning resources and accounting for the executor's resources in every task, instead of just in the first one.

@brndnmtthws Am I correct in assuming `subtractedExecutorResources` does essentially the same thing after the PR #65?

erikdw: Ah, thanks for pointing that out, definitely a screw up by me in removing the assignment to `subtractedExecutorResources` there and not reintroducing it in the place it used to be set to true.

erikdw: Fixed:  https://github.com/mesos/storm/pull/79/files

brndnmtthws: Yes, that looks right.

erikdw: Soo... my fix is buggy.  The 2nd time through the loop where `subtractedExecutorResources` is `true`, it won't enter the block that sets the `executorCpuResources` and `executorMemResources` variables, which causes an NPE further down when building the `ExecutorInfo`.

The immediate fix is simple, just set those variables unconditionally... but that led me to read [the logviewer port-acquisition logic](https://github.com/mesos/storm/blob/7333f9e02a13a747c50f53992831630f7cd1eea6/src/main/storm/mesos/MesosNimbus.java#L759-L770).  I'm pretty sure it has multiple issues.  Gotta test more...

brndnmtthws: Maybe we can write a unit test to validate the behaviour? And also catch regressions.

erikdw: That's definitely the ideal.  I'll look into how to adjust the unit tests.

PR #79, opened 2015-12-04
2fbd817 incorrectly completely removed the setting of subtractedExecutorResources to true, reintroducing it in a more natural location

erikdw:


JessicaLHartog:
:+1:

PR #80, opened 2015-12-08
Improve vagrant setup

erikdw:
1. Don't port forward to invalid ports (above 65535)
2. Don't port forward the mesos-slave and mesos-master ports, since those daemons only listen on the host-only adapters.
3. Forward port 6627 which is the Storm Nimbus thrift port, in case you want to submit topologies by making a connection from outside the master VM, and by using localhost:6627 instead of the host-only-master-IP:6627.
4. Avoid adding duplicate entries to /etc/hosts when repeatedly provisioning or running the start scripts.
5. Rename vagrant-launched start scripts to be a bit clearer on the intent of each.
6. Add date to the "starting" log for each service.
7. Instead of just stopping the mesos-master on the slave VM, we also disable upstart's automatic starting of mesos-master on the slave VM.
8. Remove incorrect use of sudo with echo; common problem in docs and provisioning shell scripts.  The provision scripts are run under sudo by vagrant by default. When run manually, they must be run under sudo as well.
9. Allow reprovisioning to succeed by changing an element of the inline provision script to use `ln -sf` (force) instead of just `ln -s`
10. Don't pass IP to start-nimbus.sh script, since it doesn't use it.
11. Set IPs in variables instead of hard-coding them.
12. Add a script which can be used to submit a topology JAR file, and document it.
13. Link from base repo document to Vagrant docs.
14. Removed trailing whitespace.
15. Removed tab chars.
16. Make apt-get provisioning invocations "loggable" by suppressing progress indicator ('-q' option)
17. Prevent "default: stdin: is not a tty" error during vagrant provisioning.
18. Enable running of docker images by enabling the containerizer in the mesos-slave configs set by the vagrant-launched scripts.
19. Prevent problem with MesosNimbus failing to start because of being unable to create its configured log dir (config: `storm.log.dir` => `/vagrant/_release/storm-mesos-0.9.6/logs`).  This is done by creating the "logs" dir in the release dir during the `build-release.sh` packaging step.
  * root cause: seemingly a bug in the virtualbox guest additions file sharing driver: if a dir previously accessed by the guest is removed from the host, then the guest gets confused and thinks the dir is present even though it really isn't anymore
    * `mkdir` within guest failing because "the file exists", even though `ls` (on outer host or within guest) shows that the file doesn't exist
20. improve build by downloading storm binary package in tarball format instead of zip, so that we have better perms in resultant package.
21. improve build by setting user & group IDs as 0 instead of whoever happens to build it
22. improve comments in storm.yaml
23. made the 2nd VM ("slave") optional, and have it off by default, since the 1st VM ("master") has all the components needed (mesos-master, mesos-slave, mesos-nimbus, storm ui, zookeeper).
24. during provision script runs, stop/kill previous storm & mesos daemons. Really we should be doing some more formal starting of the nimbus & ui a la runit or something, but that can be fixed later.
25. set the ownership of the files in the tarball package to be root.
26. check that tar is gnu-tar and not bsdtar (e.g., --owner should be supported)
27. document location of mesos configs
28. document location of zookeeper configs & data
29. added links to vagrant.md to the Web UIs of mesos and storm.
30. added info to vagrant.md docs about where various log files of interest can be found.
31. wipe out mesos-slave work_dir on each vagrant provision, to allow ease of accessing the executor & task logs via wildcards.
32. Adjust bin/build-release.sh dockerImage function to just run the docker build, which also runs bin/build-release.sh from within the docker container.


docs/vagrant.md
```diff
@@ -7,25 +7,155 @@ You can use following guide to setup a cluster in a virtual machine.
...
 
 ### Preparation
-* bin/build-release.sh must rename _release/apache-storm-${RELEASE} to _release/storm-mesos-${RELEASE}
-* sudo echo "192.168.50.101 master" >> /etc/hosts
-* sudo echo "192.168.50.102 slave"  >> /etc/hosts
+1. Add name lookup entries for the 2 VMs:
+
+  ```shell
+  sudo sh -c 'echo "192.168.50.101 master" >> /etc/hosts'
+  sudo sh -c 'echo "192.168.50.102 slave"  >> /etc/hosts'
```
JessicaLHartog: Are we keeping this naming convention? Or should we [preemptively switch](https://issues.apache.org/jira/browse/MESOS-1478) to `agent`?

docs/vagrant.md
```diff
@@ -7,25 +7,155 @@ You can use following guide to setup a cluster in a virtual machine.
...
+
+### Cluster Web Interfaces
+
+* [Storm Nimbus UI](http://master:8080)
+* [Mesos UI](http://master:5050)
+
+### Debugging:
+WARNING: If you do not correctly configure `storm.yaml` by setting all of the mandatory options specified in this project's README.md, then various things will break, and it may be hard to figure out exactly what is at fault.
+
+Various logs can be access from within the VM:
```
JessicaLHartog: accessed

docs/vagrant.md
```diff
@@ -7,25 +7,155 @@ You can use following guide to setup a cluster in a virtual machine.
...
+# mesos-master
+/var/log/mesos/mesos-master.INFO
+
+# mesos-slave
+/var/log/mesos/mesos-slave.INFO
+
+# For below mesos-executor related logs, replace the following variables:
+#  SLAVE_ID - the ID of the mesos-slave instance
+#  FRAMEWORK_ID - the ID of the framework you are running.  Just use the newest dir in that location.
+#  TOPOLOGY_ID - the ID of the storm topology you are running
```
JessicaLHartog: It might be worth mentioning that this is the ID that Storm reports in its UI, not simply the name of the topology that you submitted.

docs/vagrant.md
erikdw: For now, let's leave it be.  Easy to change later when the `agent` rename *actually* starts rolling out.

docs/vagrant.md
erikdw: Yup, this is the *ID* not the name.  I can add a clarification.

docs/vagrant.md
erikdw: thanks

docs/vagrant.md
JessicaLHartog: Sounds good. I just figured that while we were mucking around with the vagrant setup it would be an easy change to make. But it makes sense to wait until the compatibility-breaking release version of Mesos before we make this change.

docs/vagrant.md
```diff
@@ -7,25 +7,155 @@ You can use following guide to setup a cluster in a virtual machine.
...
 bin/build-release.sh
-cp vagrant/Vagrantfile .
 ```
 
 To start the cluster run following:
 ```shell
 vagrant up
 ```
-At this point the VM will have two node mesos cluster running, node 1 (master) will have a mesos master and slave running and node 2 (slave) will have only a mesos slave running. Docker has been installed as well.
+This will bring up a 2-node mesos cluster:
```
choang: this doesn't seem to be true.  Here's the code:
```
  if enable_second_slave
    config.vm.define "slave" do |node|
      node.vm.box = "trusty64-vm-2"
      node.vm.hostname = "slave"
      node.vm.network :private_network, ip: slave_ip
      node.vm.provision "shell", path: "vagrant/start-mesos-slave.sh", args: [master_ip, slave_ip]
    end
  end
```
and a bit earlier in the Vagrantfile we have:
```
Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
  master_ip = "192.168.50.101"
  slave_ip  = "192.168.50.102"
  # By default we only launch the 1st VM ("master") which has all the needed components:
  #  zookeeper, mesos-master, mesos-slave, storm UI, mesos-nimbus
  # However, you may want to enable the 2nd VM to better emulate a real cluster.
  enable_second_slave = false
```

erikdw: thanks, documenting that.

erikdw:
@choang & @JessicaLHartog : I've pushed some fixes for your comments.

choang:
I can confirm that the `vagrant up` workflow works flawlessly.  :ship: for that piece.  (I saw references to Docker which I did not test)

erikdw:
@choang : thanks for testing & the review. The docker stuff is not fully "in-scope" for this change really, I just wanted to make it clear that `bin/build-release.sh dockerImage` was broken by #65's changes to the Dockerfile, and thus simply replaced its contents with a trivial invocation of a docker-build.  It should probably be adjusted, will address later.

erikdw:
@choang & @JessicaLHartog & @dskarthick:  I followed Chi's tip for passing the storm.yaml into the MesosNimbus, it works great!  See setting of `STORM_CONF_DIR` in `vagrant/start-nimbus.sh`.

Was able to show that we can simply set `storm.log.dir` in `storm.yaml` in the MesosNimbus, and it will propagate to the MesosSupervisor & workers, thus obviating any need for us to be modifying `cluster.xml` explicitly.  Woot.

JessicaLHartog:
It looks like the changes here will have several conflicts with those in #77. Perhaps we can :ship: this one sooner rather than later, so that #77 can re-base and that PR will become easier/shorter to review?

Issue #81, opened 2015-12-10
add unit test for more than 1 storm workers being launched in 1 mesos executor


erikdw:
We currently lack unit tests for this situation.  As a result we've made an ugly chain of changes:

PR #65 created a subtle regression where we could never launch more than 1 worker in a given executor (so at most 1 worker process per topology per host).
PR #78 attempted to fix this issue, but also added a new bug, where we subtract executor resources from offers more than necessary.
PR #79 attempted to fix the bug introduced in PR #78, but it introduced another bug (a NullPointerException crash, where after the first pass through the `computeResourcesForSlot()` main loop, the `executorCpuResources` and `executorMemResources` variables aren't set to null and didn't get populated because `subtractedExecutorResources ` was already true on the 1st pass).

So we should add a unit test that prevents these kinds of regressions.

PR #82, opened 2015-12-10
Fix another regression: a crash introduced in commit feb892f

erikdw:
Commit feb892f (from PR #79) introduced a NullPointerException crash:

After the first pass through the `computeResourcesForSlot()` main loop, the
`executorCpuResources` and `executorMemResources` variables are not populated,
and thus incorreclty null, because `subtractedExecutorResources ` was set to
true on the 1st pass.

The fix is to unconditionally set them to the requested resource amounts
unconditionally, and only subtract the resources out of the offer on the
1st pass.

Also an unrelated change:  clarify that the reservations that are being skipped
in a couple functions are *dynamic* reservations.

### DIFF
--- a/src/main/storm/mesos/MesosNimbus.java
+++ b/src/main/storm/mesos/MesosNimbus.java
@@ -306,7 +306,7 @@ public class MesosNimbus implements INimbus {
 
     for (Resource r : offer.getResourcesList()) {
       if (r.hasReservation()) {
-        // skip resources with reservations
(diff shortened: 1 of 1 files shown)


erikdw:
@jessicalhartog: please review.

JessicaLHartog:
LGTM :+1:

erikdw:
Here goes nothing!  ;-)   Hope this works well for everyone.  Definitely need to fix #81

PR #84, opened 2015-12-12
Shade all possible dependencies except mesos. 

dsKarthick:
The dependencies in the mesos/storm could potentially conflict with those in the topologies. For instance, mesos/storm codebase depends on guava-18. If the topology submitted by the users of the mesos/storm cluster depend on a version of guava that is not compatible with guava-18, it would result in a conflict. So its good practice to shade all the dependencies of mesos/storm.
 
We aren't able to shade org.apache.mesos because it is loaded directly by the Mesos
"drivers", that run within the MesosNimbus and MesosSupervisor processes.

@erikdw @JessicaLHartog @brndnmtthws @tnachen @JessicaLHartog

### DIFF
--- a/bin/build-release.sh
+++ b/bin/build-release.sh
@@ -69,6 +69,10 @@ function package {(
   _rm $stormDir/*.jar
 
   # copies storm-mesos jar over
+  # We only want the shaded jar. Its important to remove the original
+  # jar so we dont have both shaded as well as original jar in the classpath
+  # for mesos nimbus
+  rm target/original-storm-0.9.6.jar
   cp target/*.jar $stormDir/lib/
   cp bin/storm-mesos $stormDir/bin/
   cp bin/run-with-marathon.sh $stormDir/bin/
--- a/pom.xml
+++ b/pom.xml
@@ -61,7 +61,50 @@
                 <exclude>defaults.yaml</exclude>
               </excludes>
             </filter>
-          </filters>
+	  </filters>
+	  <!-- The mesos/storm dependencies can conflict with the topology jar's dependencies.
+               To avoid this problem, we shade every dependency of mesos/storm that we can.  We weren't
+	       able to shade org.apache.mesos because it is loaded directly by the Mesos "drivers",
+               that run within the MesosNimbus and MesosSupervisor processes.
+         -->
+          <relocations>
+            <relocation>
+              <pattern>com.google.common</pattern>
+              <shadedPattern>com.google.common.shaded</shadedPattern>
+            </relocation>
+            <relocation>
+              <pattern>com.google.protobuf</pattern>
+              <shadedPattern>com.google.protobuf.shaded</shadedPattern>
+            </relocation>
+	    <relocation>
+              <pattern>org.apache.commons</pattern>
+              <shadedPattern>com.apache.commons.shaded</shadedPattern>
+            </relocation>
+            <relocation>
+	      <pattern>org.slf4j</pattern>
+              <shadedPattern>org.slf4j.shaded</shadedPattern>
+            </relocation>
+            <relocation>
+	      <pattern>org.apache.log4j</pattern>
+              <shadedPattern>org.apache.log4j.shaded</shadedPattern>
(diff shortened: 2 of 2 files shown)


brndnmtthws:
Makes sense! Let me test it out, and I'll merge it provided everything's okay.

dsKarthick:
Thanks @brndnmtthws

dsKarthick:
@brndnmtthws With ^^ commit, you should be able to test the new build with vagrant. I submitted a topology and its working. Please let me know if you face any issues.

brndnmtthws:
There's one minor issue, which is that we need to include `commons-io` in the dependencies for the supervisor. I'll merge the other PR with your commits.

dsKarthick:
@brndnmtthws Why does shading force us to include `commons-io`?

brndnmtthws:
Because the scheduler requires it, and it's included elsewhere in the classpath but not under the name we're shading `org.apache.commons` to.

dsKarthick:
@brndnmtthws Mind explaining a bit more? Are we talking about [https://github.com/mesos/storm/blob/master/src/main/storm/mesos/NimbusScheduler.java](NimbusScheduler.java)

brndnmtthws:
Specifically this line of code (IIRC): https://github.com/mesos/storm/blob/master/src/main/storm/mesos/logviewer/SocketUrlDetection.java#L48

dsKarthick:
@brndnmtthws Line you referenced should work without including `org.apache.commons`

erikdw:
@dsKarthick : I too am curious about what spurred @brndnmtthws to notice this implicit dependency and then explicitly include it.  The `storm-core` dependency being a provided dependency of this project already implicitly brought along `commons-io` (see `mvn dependency:tree -Dverbose`).  And `commons-io-2.4.jar` is one of those "provided in the storm classpath" jars that comes in the binary storm release.

So yes, @dsKarthick it should not be strictly necessary to include this `commons-io` dependency explicitly.  But the dependency *does* exist and I don't think it hurts to be explicit here and also lock in our version of the dependency.  Protects us against future shading within the core storm code, which I believe in `storm-0.10.0` has already shaded away `commons-io`.  I think that would break this project without @brndnmtthws 's changes.   That being said I'd still appreciate an explicit explanation for how this implicit dependency was noticed, given that it shouldn't cause issues in `storm-0.9.6` as far as I can tell.

erikdw:
Ohh... or (upon closer reading of one of @brndnmtthws responses to you @dsKarthick ) maybe it's that the logviewer detection code dies in the supervisor process without that dependency inclusion, since the `commons-io-2.4.jar` sitting in the classpath doesn't have the right package path (`org.apache.commons.shaded.io.IOUtils` vs. `org.apache.commons.io.IOUtils`), so it cannot be loaded by the supervisor's JVM, thus causing a failure of that process.

erikdw:
Yes, just talked this over with @dsKarthick :  

The bits in the storm-mesos framework will have been massaged so that it expects to find the class at `org.apache.commons.shaded.io.IOUtils`, but the class doesn't exist at that fully-qualified classname in the bundled-up/shaded storm-mesos jar, nor does it exist in the storm-provided jars on the worker host (`commons-io-2.4.jar` has `org.apache.commons.io.IOUtils` not `org.apache.commons.shaded.io.IOUtils`).

More specifically, yes, the shaded storm-mesos jar *did not* have the `commons-io` classes in it at all, because `commons-io` was a *provided* dependency through the inclusion of the *provided* `storm-core` dependency, and so was assumed to exist in the target environment.  But now it didn't anymore after the shading, because of the mismatched fully-qualified classnames.  So @brndnmtthws 's change to make `commons-io` a *compile* scoped dependency has pulled those classes into the shaded jar and thus the class will be found at the expected fully-qualified classname.

Phew.  So I think we're all good here now.

dsKarthick:
@erikdw, Thanks for the detailed responses. Just as a best practice, in the future, its good to explicitly list any dependencies we use directly in this repository.

PR #85, opened 2015-12-13
fix bug in the UI regarding how it displays the number of slots in the cluster

debugger87:
The MesosSupervisor has not implemented `getMetadata`, so that the total slots is always zero. I just implement is using the `_stats.snapshot()` to create a `PersistentVector` which contains the ports of workers.


debugger87:
@jessicalhartog: please review.

debugger87:
@erikdw please review.

brndnmtthws:
Cool, thanks! Tested & verified.

PR #87, opened 2016-01-06
Changes to get docker working on vagrant.

dsKarthick:
Detailed list of changes:
  - Get rid of -q parameter in apt-get. No harm printing extra info -
   infact, we want to know more detailed information about errors, if
   any.
 - Set a lower timeout value for apt-get. This helps avoid long pause
   times when "apt-get update" is run thereby improving the time that it
   takes to bring up the vagrant machine.
 - Fix the broken "bin/build-release.sh dockerImage" functionality
 - Tune the memory settings for supervisor and worker process in order
   to avoid container from exiting due to OOM
 - Set MESOS_NATIVE_JAVA_LIBRARY environment variable while launching
   nimbus inorder to avoid container failing like below
~~~
     # vagrant@master:~$ sudo docker logs 84ad44355638
     #  SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
     #  SLF4J: Defaulting to no-operation (NOP) logger implementation
     #  SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder
     #  for further details.
     #  Failed to load native Mesos library from null
~~~

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -5,6 +5,12 @@ VAGRANTFILE_API_VERSION = "2"
 
 $provision_script = <<SCRIPT
 
+function install_package {
+  name_of_package=$1
+  echo "Installing ${name_of_package}"
+  apt-get -o Acquire::http::Timeout=1 -o Acquire::ftp::Timeout=1 -y install ${name_of_package}
+}
+
 PREFIX="PROVISIONER:"
 
 set -e
@@ -20,16 +26,22 @@ DISTRO=$(lsb_release -is | tr '[:upper:]' '[:lower:]')
 CODENAME=$(lsb_release -cs)
 echo "deb http://repos.mesosphere.io/${DISTRO} ${CODENAME} main" | sudo tee /etc/apt/sources.list.d/mesosphere.list
 
-apt-get -q -y update
+apt-get -o Acquire::http::Timeout=1 -o Acquire::ftp::Timeout=1 -y update
 echo oracle-java8-installer shared/accepted-oracle-license-v1-1 select true | sudo /usr/bin/debconf-set-selections
-apt-get -q -y install oracle-java8-installer
-apt-get -q -y install oracle-java8-set-default
-apt-get -q -y install libcurl3
-apt-get -q -y install zookeeperd
-apt-get -q -y install aria2
+
+echo "Installing oracle-java8-installer..."
+install_package "oracle-java8-installer"
+echo "Installing oracle-java8-set-default..."
+apt-get -o Acquire::http::Timeout=1 -o Acquire::ftp::Timeout=1 -y install oracle-java8-set-default
+echo "Installing libcurl3..."
+apt-get -o Acquire::http::Timeout=1 -o Acquire::ftp::Timeout=1 -y install libcurl3
+echo "Installing zookeeperd..."
+apt-get -o Acquire::http::Timeout=1 -o Acquire::ftp::Timeout=1 -y install zookeeperd
+echo "Installing aria2..."
+apt-get -o Acquire::http::Timeout=1 -o Acquire::ftp::Timeout=1 -y install aria2
 
 echo "${PREFIX}Installing mesos ..."
-apt-get -q -y install mesos
+apt-get -o Acquire::http::Timeout=1 -o Acquire::ftp::Timeout=1 -y install mesos=0.25.0-0.2.70.ubuntu1404
 echo "Done"
 
 ln -sf /usr/lib/jvm/java-8-oracle/jre/lib/amd64/server/libjvm.so /usr/lib/libjvm.so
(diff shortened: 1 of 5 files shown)


src/main/storm/mesos/MesosNimbus.java
```diff
@@ -890,7 +890,7 @@ private Credential getCredential(FrameworkInfo.Builder finfo) {
       if (StringUtils.isNotEmpty(secretFilename)) {
         try {
           // The secret cannot have a NewLine after it
-          credentialBuilder.setSecret(ByteString.readFrom(new FileInputStream(secretFilename)));
```
erikdw: @DarinJ is changing this in a different way (using reflection) to support both mesos-0.26.0 and earlier versions.

Vagrantfile
```diff
@@ -20,16 +20,22 @@ DISTRO=$(lsb_release -is | tr '[:upper:]' '[:lower:]')
 CODENAME=$(lsb_release -cs)
 echo "deb http://repos.mesosphere.io/${DISTRO} ${CODENAME} main" | sudo tee /etc/apt/sources.list.d/mesosphere.list
 
-apt-get -q -y update
+apt-get -o Acquire::http::Timeout=1 -o Acquire::ftp::Timeout=1 -y update
```
erikdw: You mention this timeout change in the description:   lower timeout value for apt-get. This helps avoid long pause times when "apt-get update" is run thereby improving the time that it takes to build a docker image."

Is it safe/correct to do that?  i.e., if the timeout occurs then presumably it will break the resultant image?

Vagrantfile
```diff
@@ -20,16 +20,22 @@ DISTRO=$(lsb_release -is | tr '[:upper:]' '[:lower:]')
...
 echo oracle-java8-installer shared/accepted-oracle-license-v1-1 select true | sudo /usr/bin/debconf-set-selections
-apt-get -q -y install oracle-java8-installer
-apt-get -q -y install oracle-java8-set-default
-apt-get -q -y install libcurl3
-apt-get -q -y install zookeeperd
-apt-get -q -y install aria2
```
erikdw: The reason I put `-q` in is to suppress complaints about stuff already existing when I reprovision (as you do when you iteratively develop).
Perhaps instead of just removing `-q` we should instead check for the existence of the installed package, and if it isn't installed we proceed to do the package installation without `-q`.

src/main/storm/mesos/MesosNimbus.java
DarinJ: I went ahead and pushed that change so you can get the idea.

Vagrantfile
dsKarthick: @erikdw There aren't so many error messages when the package is aready installed. Following is what is displayed when oracle-java8-installer is already installed

```
==> master: Reading package lists...
==> master: Installing oracle-java8-installer...
==> master: Installing oracle-java8-installer
==> master: Reading package lists...
==> master: Building dependency tree...
==> master: Reading state information...
==> master: oracle-java8-installer is already the newest version.
==> master: 0 upgraded, 0 newly installed, 0 to remove and 31 not upgraded.
==> master: Installing oracle-java8-set-default...
```

Vagrantfile
dsKarthick: @erikdw You are right. This change will help speed up `vagrant up` not building docker Image as I previously stated. In terms of whether or not its dangerous, we should specify the version explicitly if we think we need it for mesos-storm. If that version cannot be found, then ` vagrant up` should error out. As per your recommendation, I will specify the specific version of mesos for installation.

src/main/storm/mesos/MesosNimbus.java
dsKarthick: Thanks @erikdw for pointing out and thanks @DarinJ for pushing the changing. I will get rid of this change.

dsKarthick:
@erikdw :eyes:

storm.yaml
```diff
@@ -30,11 +30,11 @@ supervisor.autostart.logviewer: true
 # Use the public Mesosphere Storm build
 # Please note that it won't work with other distributions.
 # You may want to make this empty if you use `mesos.container.docker.image` instead.
-# mesos.executor.uri: "http://downloads.mesosphere.io/storm/storm-mesos-0.9.0.1.tgz"
+# mesos.executor.uri: "file:///usr/local/storm/storm-mesos-0.9.6.tgz"
 
 # Alternatively, use a Docker image instead of URI. If an image is specified,
 # Docker will be used instead of Mesos containers.
-mesos.container.docker.image: mesosphere/storm
+mesos.container.docker.image: "karthick/mesos-storm:git-f7ae208"
```
erikdw: you sure you wanna commit this change?

storm.yaml
```diff
@@ -30,11 +30,11 @@ supervisor.autostart.logviewer: true
 # Use the public Mesosphere Storm build
 # Please note that it won't work with other distributions.
 # You may want to make this empty if you use `mesos.container.docker.image` instead.
-# mesos.executor.uri: "http://downloads.mesosphere.io/storm/storm-mesos-0.9.0.1.tgz"
+# mesos.executor.uri: "file:///usr/local/storm/storm-mesos-0.9.6.tgz"
```
erikdw: I think this is ok -- we want people to be adjusting this themselves for now.

erikdw:
LGTM @dsKarthick !   After you squash I'll merge.

erikdw:
thanks @dsKarthick, merging now

dsKarthick:
thanks for merging @erikdw

PR #88, opened 2016-01-16
Fix NoClassDefFoundError with shading

mcoffin:
This fixes a `NoClassDefFoundError` caused by the shading that we do.

`MesosNimbus` explicitly uses the non-lang3 version of `StringUtils` from `commons-lang`, but it wasn't being included in the shaded jar, but the namespace filter `org.apache.commons` was picking it out regardless.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -179,6 +179,11 @@
       <version>3.4</version>
     </dependency>
     <dependency>
+      <groupId>commons-lang</groupId>
+      <artifactId>commons-lang</artifactId>
+      <version>2.5</version>
+    </dependency>
+    <dependency>
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
       <version>2.4</version>


erikdw:
@mcoffin : good catch!  This should *definitely* be included explicitly if we are relying on it. [1] 
However, instead of adding the `commons-lang 2.5` dependency maybe we can just change MesosNimbus.java to use the lang3 version of commons-lang?

[1] I'm wondering if there is a foolproof easy way to check that all dependencies are included explicitly instead of inherited from those provided by Storm's bundled jars.

DarinJ:
I updated MesosNimbus to lang3 as part of PR #77, as it solved a Storm 0.10.0/0.9.6 compatibility issue easily.  We may want to see if adding the dependency is still needed.

erikdw:
@DarinJ : ahh, nice.  Yeah, I agree that with your change in PR #77 we shouldn't need this fix.

mcoffin:
> with [the] change in PR #77 we shouldn't need this fix.

Quick search for other references to regular commons-lang agrees. I'll close when #77 lands.

erikdw:
@mcoffin : I'm gonna close this now since I finally merged in @DarinJ's #77 which has the regular commons-lang, thanks for your contribution.

Issue #83, opened 2015-12-10
automate release tagging and building


erikdw:
We will soon have an actual versioning scheme implemented in this project (see #18, #76, #77).
When #77 is done, we can follow the process documented in #18 to perform releases.

However, it would be way snazzier if we could leverage travis-ci.org to:
1. automatically build the framework artifacts/packages/tarballs
2. tag the code with the version
3. upload the resultant framework packages/tarballs into the github repo

salimane:
@erikdw  i could take of this.

erikdw:
@salimane : that would be great -- hopefully you've followed the discussions in #109 since it is related.

salimane:
@erikdw yep  and I think it's better to start somewhere. The first goal is to magically have these on the releases page:

```
storm-mesos-0.1.0-storm0.10.0-mesos0.28.0.tgz
storm-mesos-0.1.0-storm0.10.0-mesos0.27.1.tgz
storm-mesos-0.1.0-storm0.9.6-mesos0.28.0.tgz
storm-mesos-0.1.0-storm0.9.6-mesos0.27.1.tgz
```

Then, on docker hub, magically have these docker images appear : 
```
mesos/storm-mesos:0.1.0-0.10.0-0.28.0
mesos/storm-mesos:0.1.0-0.10.0-0.27.1
mesos/storm-mesos:0.1.0-0.9.6-0.28.0
mesos/storm-mesos:0.1.0-0.9.6-0.27.1
```

Thanks

erikdw:
@salimane : sounds great.  FYI, you *should* be able to use your own fork of this project for testing the auto-tagging-and-releasing.  TravisCI provides free service to OSS repos.  Not sure about DockerHub.

salimane:
@erikdw yep true

erikdw:
Interestingly, when I merged #117, it automatically closed this issue #83.  That is because the commit message contains the words: `fixes #83`

See this for more info:
 * https://help.github.com/articles/closing-issues-via-commit-messages/

Issue #18, opened 2014-11-05
Download links for storm-mesos releases


asanami:
Can you please provide clarification on where to pick up releases for storm-mesos.  The storm.yaml file references download links to:

http://downloads.mesosphere.io/storm/storm-mesos-0.9.0.1.tgz

I cannot find releases from the official mesosphere web page, and more specifically looking for the storm-mesos-0.9.2 release.

Thanks

erikdw:
The plan is to have these releases available for download from within the github project page.

But before that we need some version scheme.  i.e., the framework has enough changes that simply referring to the storm version is insufficient.   We should have a version for the framework itself, and also reference the storm version that is included within the package.  e.g., like how kafka packages reference the scala version they are built with:
* `kafka_2.9.1-0.8.2.2` == Kafka Version `0.8.2.2` built with Scala version `2.9.1`

erikdw:
The version naming scheme is being discussed in #76

erikdw:
Per #76, we've agreed upon a version naming scheme, so when PR #77 is complete we can proceed publish our first "official" releases on this github project page.

### How to upload releases
* https://github.com/blog/1547-release-your-software
* https://help.github.com/articles/about-releases/
  * notably, releases are allowed to be up to 2GB in size, so we're good with the ~26MB releases we currently have

### Building releases
I'm not sure of the full process yet, but here's a stab at it:

#### dreary present: super manual process
1. manually rev the version
2. `bin/build-release.sh`
3. gci -m "release version X.Y.Z"; git tag vX.Y.Z; git push; git push --tags
4. Upload the tarball to this [project's releases view](https://github.com/mesos/storm/releases)

#### copacetic short-term improvement: less manual process with maven assistance
Then we can move to leverage maven for automating some of the painful manual steps in there:
1. run the `maven assembly` plugin (or some other one) for running `bin/build-release.sh`
2. use the `maven release` plugin to manage the versions updating and tagging
3. we still manually upload the tarball to this [project's releases view](https://github.com/mesos/storm/releases)

#### glorious future: full automation

Ideally we'd leverage travis-ci.org.  See #83.

erikdw:
Now that @DarinJ's #77 change has landed we have the basis for creating actual builds.  I'll spend some time building a few variations and uploading them to the github project's releases page.

erikdw:
Awesomely we have reached the "glorious future" already with @salimane's contribution in #117.

[...]

PR #91, opened 2016-02-02
Pass env vars to docker build

### LINKED ISSUE #90: bin/build-release.sh dockerImage should be passed environment variables
When I run `STORM_RELEASE=0.10.0 bin/build-release.sh dockerImage` the docker build command that gets run ignores my `STORM_RELEASE` setting when it runs.

drewrobb:
@erikdw fixes #90, although a little bit janky because of repeating variable names 3 times...

Also, perhaps the base image should match MESOS_RELEASE, but that is a separate issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,10 @@ ENV JAVA_HOME /usr/lib/jvm/java-7-openjdk-amd64
 
 ENV MESOS_NATIVE_JAVA_LIBRARY /usr/lib/libmesos.so
 
+ARG MESOS_RELEASE=0.25.0
+ARG STORM_RELEASE=0.9.6
+ARG MIRROR=http://www.gtlib.gatech.edu/pub
+
 ADD . /work
 
 WORKDIR /work
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 #
 # Dockerfile for Storm Mesos framework
 #
-FROM mesosphere/mesos:0.25.0-0.2.70.ubuntu1404
+FROM mesosphere/mesos:0.27.0-0.2.190.ubuntu1404
 MAINTAINER Timothy Chen <tnachen@gmail.com>
 
 ENV DEBIAN_FRONTEND noninteractive
@@ -11,7 +11,7 @@ ENV JAVA_HOME /usr/lib/jvm/java-7-openjdk-amd64
 
 ENV MESOS_NATIVE_JAVA_LIBRARY /usr/lib/libmesos.so
 
-ARG MESOS_RELEASE=0.25.0
+ARG MESOS_RELEASE=0.27.0
 ARG STORM_RELEASE=0.9.6
 ARG MIRROR=http://www.gtlib.gatech.edu/pub
 
--- a/bin/build-release.sh
+++ b/bin/build-release.sh
@@ -109,7 +109,11 @@ function package {(
 )}
 
 function dockerImage {(
-  cmd="docker build -t mesos/storm:git-`git rev-parse --short HEAD` ."
+  cmd="docker build \
+       --build-arg MESOS_RELEASE=$MESOS_RELEASE \
+       --build-arg STORM_RELEASE=$STORM_RELEASE \
+       --build-arg MIRROR=$MIRROR \
+        -t mesos/storm:git-`git rev-parse --short HEAD` ."
   echo $cmd
   $cmd
 )}
--- a/pom.xml
+++ b/pom.xml
@@ -28,8 +28,8 @@
     <maven.compiler.source>1.7</maven.compiler.source>
     <maven.compiler.target>1.7</maven.compiler.target>
     <storm.default.version>0.9.6</storm.default.version>
-    <mesos.default.version>0.24.1</mesos.default.version>
-    <mesos.version>0.24.1</mesos.version>
+    <mesos.default.version>0.27.0</mesos.default.version>
+    <mesos.version>0.27.0</mesos.version>
   </properties>
 
   <profiles>


dsKarthick:
@drewrobb Fyi, here is an alternative way of passing environment variable https://github.com/docker/docker/issues/2637#issuecomment-64827330 . That said, I prefer your approach.

drewrobb:
@dsKarthick cool trick!

erikdw:
@drewrobb & @dsKarthick : it may actually be better to use that env file trick to integrate with docker hub.  I'm not sure of how the dockerhub mechanism works, and it's unclear to me how we'll set these parameters when using dockerhub.  Maybe they're unneeded.

I'm a bit worried that the proposed change here in #91 might not work if we don't have all those variables specified.  ...   Oh!  You can specify a default value:
* https://docs.docker.com/engine/reference/builder/#arg

```
ARG <name>[=<default value>]
```

That should suffice until we can gain a better understanding of the docker hub integration.  Maybe @brndnmtthws can speak to that?

erikdw:
@DarinJ just created an issue (#92) that is directly related to this.

Dockerfile
```diff
@@ -11,6 +11,10 @@ ENV JAVA_HOME /usr/lib/jvm/java-7-openjdk-amd64
 
 ENV MESOS_NATIVE_JAVA_LIBRARY /usr/lib/libmesos.so
 
+ARG MESOS_RELEASE
+ARG STORM_RELEASE
+ARG MIRROR
```
dsKarthick: @drewrobb Lets set the default values for these variables like @erikdw mentioned?

dsKarthick:
@erikdw Good point regarding the default values.

Dockerfile
drewrobb: There are default values set in bin/build-release.sh (since the environment variables are always set to something, so I think it would just be making those values redundantly specified unless I'm missing something?

dsKarthick: What happens when someone tries to do `docker build -t <tag_name> ` ?

drewrobb: Ah good point, I was thinking people would only invoke build via the build-release.sh script.

dsKarthick:
@drewrobb @erikdw On a second thought, I feel a combination of sourcing technique and @drewrobb's build-arg technique would be better for couple of reasons
      
      - It allows us to specify as many environment variables as we like in the file without making $cmd in build-release.sh look big
      - It helps keep the Dockerfile small as well.

Any thoughts?

bin/build-release.sh
```diff
@@ -109,7 +109,11 @@ function package {(
 )}
 
 function dockerImage {(
-  cmd="docker build -t mesos/storm:git-`git rev-parse --short HEAD` ."
+  cmd="docker build \
+       --build-arg MESOS_RELEASE=$MESOS_RELEASE \
+       --build-arg STORM_RELEASE=$STORM_RELEASE \
+       --build-arg MIRROR=$MIRROR \
+        -t mesos/storm:git-`git rev-parse --short HEAD` ."
```
DarinJ: Would be cool here to embed the storm/mesos versions into the tag (possibly in the version).  Then it'd be trivial to release a few different images on dockerhub.  Related to #18, #83.

DarinJ:
Might consider using a stock Ubuntu image and installing the correct version of mesos as defined by `MESOS_VERSION`.  Debs available here: https://open.mesosphere.com/downloads/mesos/

brndnmtthws:
:+1: for parameterizing the version. I'm not sure how to make it integrate well with the automated builds on Docker hub, however.

erikdw:
Ok, so the goals seem to be:  

1. Allow the `bin/build-release.sh dockerImage` invocation to accept configurable storm & mesos versions, and properly build the docker image with the chosen version of mesos.
2. Allow the integration between GitHub & Docker Hub to continue working.

I think we can accomplish both of those with a combination of our ideas from this PR's conversation:

1.  Populate our `Dockerfile` with `ARGS` for the mesos & storm versions, giving them default values which will be used when we don't specify the `--build-arg` parameters (i.e., when Docker Hub does the build upon a commit to the mesos/storm repo).
2. Change our `Dockerfile` to be based on a vanilla ubuntu image that we install mesos on top of, to allow for the mesos version to be configurable.  i.e., DarinJ's [idea proposed above](https://github.com/mesos/storm/pull/91#issuecomment-178573510).

As an aside, something else we need to do is update the mesos dependency in the `pom.xml` to match the chosen mesos version.

Anyone see any issues with the proposal?

....

Oh, I see an issue with @DarinJ 's idea from my naive perspective:  it's not clear how we can get the actual mesos deb package name/URL given the version.  There are extra bits in the package name and URL. i.e.,
* http://downloads.mesosphere.io/master/ubuntu/15.04/mesos_0.27.0-0.2.190.ubuntu1504_amd64.deb
* http://downloads.mesosphere.io/master/ubuntu/15.04/mesos_0.26.0-0.2.145.ubuntu1504_amd64.deb
* http://downloads.mesosphere.io/master/ubuntu/15.04/mesos_0.25.0-0.2.70.ubuntu1504_amd64.deb

I don't know what `-0.2.70`, `-0.2.145`, `-0.2.190` *mean* nor how to predict them.

dsKarthick:
@erikdw Regarding point 3, what do you think about passing the mesos and storm versions as maven parameters? Currently we grep do populate https://github.com/drewrobb/mesos-storm/blob/dr/pass-env-vars/bin/build-release.sh#L12-L24 mesos and storm versions. Instead we could pass the version we intend to build like `mvn install "-DstormVersion=0.9.6 -DmesosVersion=0.26.0`?

dsKarthick:
@erikdw Also, it would be nice to source the environment variables defined in a separate file. That way we could define global env varaiables in both [here](https://github.com/mesos/storm/blob/3376813d8fa0e339c8c0321774e5572eeefb60bb/bin/build-release.sh#L80), [here](Dockerfile) and wherever else?

erikdw:
This feels like it is snowballing a bit.  I still think there are fundamental things we need to iron out to fully satisfy the goals as specified above, but it's going to take time to figure out the "right" way to handle the weird suffix thing for the mesos packages.  I've asked a [question about that](https://mail-archives.apache.org/mod_mbox/mesos-dev/201602.mbox/%3CCAO5KYW8%3DWQ2Rd1obyA6_9yzaSrDZE___Bd7g41FjgK4gcJrKFw%40mail.gmail.com%3E) on the mesos dev list.

So for now in this PR, how about we *just* do the "ARGS with defaults" thing?

dsKarthick:
@erikdw Yes. I agree. Right solutions are almost always inversely propotional to quick solutions.

drewrobb:
I like this plan. I updated my commit to include defaults (defaults that are consistent with the existing dockerfile, but not pom.xml)

erikdw:
@drewrobb : that looks good to me, let me test it later tonight or tomorrow morning, I expect I'll just merge it.

erikdw:
hey @drewrobb : can you please make 1 more change?  `pom.xml` has the mesos versions as 0.24.1, can you change them to 0.25.0 so the defaults are all the same?  Thanks!

bin/build-release.sh
dsKarthick: @DarinJ  We should also make `mesos/storm` configurable. Something like `<user_name>/<image_name>` where `<image_name>` is something that is in accordance with https://github.com/mesos/storm/issues/83

For now, lets keep that change separate so we don't block this PR on  https://github.com/mesos/storm/issues/83 ?

dsKarthick:
@drewrobb @erikdw why not mesos 0.27.0 ?

erikdw:
@dsKarthick : I'm find with `0.27.0` too, let's just be consistent in the various files where its sprinkled around for now.  I'll file an issue for follow-up work that came up in this PR.

@drewrobb : if you update the `FROM` base image, please note that you'll need to change it to `mesosphere/mesos:0.27.0-0.2.190.ubuntu1404`.

drewrobb:
Updated to mesos 0.27.0. I tested default build locally and it worked for me.

erikdw:
Thanks @drewrobb and others that contributed to the discussion.  Merging!

erikdw:
FYI, I've filed 2 issues for a the unresolved topics that came up in this review:

* #95: improve the docker image tag name
* #96: improve handling of mesos version in the build

PR #93, opened 2016-02-13
Per topology resource aware mesos-storm scheduler

dsKarthick:
In short, the pull request contains the following changes
   * A new scheduling framework
   * A DefaultScheduler that addresses starvation problem - where small topologies starve while competing for resources with big topologies
   * General code re-organization to make mesos/storm codebase pretty and less scary
   
Starvation Problem 
===============
When a big topology and a small topolog(y/ies) need resources(cpu/memory) at the same time, mesos-storm's allSlotsAvailableForScheduling calculates the number of available slots based on the requirements of the big topology. That leaves small topolog(y/ies) starving. In many cases, small topologies can easily be scheduled with the available resources eventhough big topology cannot be scheduled. 

Instead of determining the number of available slots based on the global max and min requirements, this PR works on per topology requirements to determine the number of resources available. Please look at the DefaultScheduler for more information.

Scheduling Framework
==================
Along with the fix, I created a mesos-storm scheduling framework. The new framework defines an interface named 'IMesosStormScheduler'. This allows us to implement a variety of custom schedulers that could potentially fix a lot of issues that we are currently facing in production. Following are some ideas around custom schedulers

Priority Scheduler
********************
Currently, when we restart worker nodes, all the topologies restart and try to come up at the same time. This causes a sudden flood of requirements which mesos-storm cannot satisfy - because mesos doesnt offer resources all at once but at regular intervals to mesos-storm. We could assign priorities to the topologies based on resource-requirements or business impotance. 

Locality aware Scheduler
*****************************
Storm workers generally are very chatty - there is a lot of communication going on between workers all the time. So co-locating the workers on the same nodes could be beneficial. So we could make sure that all the workers are close to each other.


dsKarthick:
@erikdw @brndnmtthws @tnachen  Please take a look when you can. 

fyi, I am planning on adding some unit tests next week. If you guys want any specific unit test that you would like to see, please feel free to comment.

brndnmtthws:
After a cursory glance, it looks good. I look forward to seeing some tests. It'd also be interesting to hear how you've tested/validate the changes.

storm/src/main/storm/mesos/schedulers/OfferResources.java
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    this.hostName = hostName;
+    this.mem = offerMem;
+    this.cpu = offerCpu;
+    if ((portRanges != null) && (!portRanges.getRangeList().isEmpty())) {
+      this.addPortRanges(portRanges);
+    }
+  }
+
```
DarinJ: Looks like you stripped out any awareness of roles here.  If a cluster has a role with reserved resources for storm, an `offer.getResourceList` will have two `Resource` of type `SCALAR` with name `cpus`, `mem`, `ports`, etc, one from the default role and one with the named role.  Here your `for` loop is simply overriding the value, so you'll get the last `Resource` of the name.

It will also complicate the TaskInfo stage a bit, as you'll need to add two resources if the total value the task needs spans the Resources of the offer.

 @brndnmtthws just added resource role awareness into Storm, I added this to Myriad a while ago as well if you have questions.

DarinJ:
Did a quick look though this and there's a lot I like, tested the build against 0.10.0 haven't ran a topology on it yet.  I think the PR broke reserved role resources though, this can be corrected easily.

dsKarthick:
@brndnmtthws @DarinJ Thanks for the feedback

@DarinJ I will work on your concern regarding role awarness

@brndnmtthws We have been running a forked version of mesos-storm with my proposed changes in production at Groupon. I am testing it on vagrant cluster plus thinking of adding some unit test coverage. How do you guys do test this at Mesosphere?

DarinJ:
@dskarthik let me know if you have questions, happy to test out on my cluster once the change is made.

storm/src/main/storm/mesos/schedulers/OfferResources.java
dsKarthick: @DarinJ In the latest commit, I have introduced a fix for skipping dynamic reservations. However I am slightly confused by you comment about your following comment

> offer.getResourceList will have two Resource of type SCALAR with name cpus, mem, ports, etc, one from the default role and one with the named role.

When I look at the current [master](https://github.com/mesos/storm/blob/master/storm/src/main/storm/mesos/MesosNimbus.java#L300-L356), I dont see any role related logic. Could you clarify your thought?

DarinJ: @dsKarthick I looked around a bit, it's sort of implicit the first relevant sections is `getResources` in [MesosNimbus.java 300-318](https://github.com/mesos/storm/blob/master/storm/src/main/storm/mesos/MesosNimbus.java#L300-L318). You replaced that method with the constructor above but note the deferences of `+=` and `=`.  By adding the values you get the full value of all Scalars with name `cpus` in the offer where one part is the will have a role the other part won't.  
The second place you see this is in `getResourceScalar` [MesosNimbus.java#L391-423](https://github.com/mesos/storm/blob/master/storm/src/main/storm/mesos/MesosNimbus.java#L391-L423), here we loop through the resources pulling out the amount needed as we find resources of the appropriate name and return a list of Resources of Type `SCALAR` and Name `cpus`.  This is important because if you have one resource with 4 cpus and another with 2, and you tried to create a task info with one Resource of name `cpus` with value 6, you'd get a task failure.  This may be worth commenting.

Anyway, I think since you're still using `getScalarResource` to create the resources, you may be OK just updating the constructor.  I'll try to review the relevant pieces of your code tomorrow to verify.

brndnmtthws: I believe @DarinJ is correct here. You should be summing the resource values across the offer, since the offer may contain resources split across roles.

This is exactly what `testGetResources()` (which you removed) was verifying.

dsKarthick: @brndnmtthws @DarinJ Got it. I will fix it first thing tomorrow and maybe add a unit test for OfferResources as a replacement to the one I deleted.

@DarinJ Nice catch!

storm/src/main/storm/mesos/schedulers/DefaultScheduler.java
```diff
@@ -0,0 +1,338 @@
...
+              if (mesosWorkerSlot == null) {
+                continue;
+              }
+              String slotId = mesosWorkerSlot.getNodeId() + ":" + mesosWorkerSlot.getPort();
+              mesosWorkerSlotMap.put(slotId, mesosWorkerSlot);
+              // Place this offer in the first bucket of the RotatingMap so that it is less likely to get rotated out
+              offers.put(resources.getOfferId(), resources.getOffer());
+              allSlots.add(mesosWorkerSlot);
+              if (--slotsNeeded == 0) {
+                break;
```
erikdw: As we found this afternoon, this should be a "double break", where we check for `slotsNeeded == 0` before continuing in the `for (Object currentNode : nodes) {` loop.   Otherwise we were setting aside too many slots for a given topology to use under certain circumstances, which resulted in an uneven spread of tasks across workers.

dsKarthick:
@erikdw @DarinJ  @brndnmtthws 

Detailed list of changes:
   * Created storm.mesos.schedulers.SchedulerUtils.java
   * Fixed bug related to supervisorExists.supervisorExists
   * Added tests for DefaultScheduler.allSlotsAvailableForScheduling
   * Created storm.mesos.schedulers.DefaultSchedulerTest.java
   * Created storm.mesos.TestUtils.java
   * Fixed a bug related to resource caculation in
     allSlotsAvailableForScheduling to account for supervisor resources
     once per node
   * Fixed a bug pointed out by Erik Weathers to break out of the
     worker slot creation loop when (slotsNeeded == 0)
   * Fixed bug pointed out by Darin where I missed to account for
     static resource reservations

TODO:
   -> Add unit tests for DefaultScheduler.schedule method
   -> Add unit tests for OfferResources

erikdw:
@dsKarthick : I think it's a bit more "standard" to break such tests out into separate functions. However, I don't have a strong opinion -- as long as it's trivial to find the broken test case from the output I'm ok with this style.  Regarding the latest change: please ensure function params have ", foo" instead of ",foo" styling (I noticed somethings that don't adhere to that -- can we add that as a checkstyle requirement?).  Also there are some superfluous empty lines at the bottom of `testAllSlotsAvailableForScheduling()`

storm/src/main/storm/mesos/util/PrettyProtobuf.java
```diff
@@ -6,62 +6,108 @@
...
- *  1. Check whether a value is set/null when deciding whether to include it.
- *  2. Currently we only include the object values that we care about for a
- *     storm-only mesos-cluster.  We should instead allow configuration for choosing
- *     which fields to include (e.g., a bitmap toggling certain fields on/off).
- *  3. Allow cleanly logging to separate Logger, to allow configuring the logs going to
- *     a separate log file.  The complication is that these methods lack context,
- *     they are just pretty-printing the protobuf objects.
+ * 1. Check whether a value is set/null when deciding whether to include it.
+ * 2. Currently we only include the object values that we care about for Groupon's
+ * storm-only mesos-cluster.  We should instead allow configuration for choosing
```
erikdw: Please re-indent as it used to be.  Removing the spacing makes it all run together into a wall of text.

storm/src/test/storm/mesos/schedulers/DefaultSchedulerTest.java
```diff
@@ -0,0 +1,351 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package storm.mesos.schedulers;
+
+/**
```
erikdw: Something is off here, duplicate License.

storm/src/main/storm/mesos/util/PrettyProtobuf.java
```diff
@@ -6,62 +6,108 @@
...
   /**
+   * Ideally we'd have a generic method for these getTrimmedString methods, but
+   * these protobuf types don't implement a common interface (which would need
+   * to include getValue()).
+   */
+
+  /**
+   * Pretty-print the mesos protobuf OfferID.
+   */
+  public static String getTrimmedString(OfferID id) {
```
erikdw: I'm not sure if we need `getTrimmedString`.  I'm pretty sure the reason it's not here in storm-mesos is that @tnachen pointed out to me in the [initial review for the logging improvements](https://github.com/mesos/storm/pull/32) that it was unnecessary.  But I left it in place in our internal implementation.  Should be tested.

storm/src/main/storm/mesos/util/PrettyProtobuf.java
```diff
@@ -176,51 +224,49 @@ private static String rangeToString(Range range) {
...
-      switch (r.getType()) {
-        case SCALAR:
-          value = String.valueOf(r.getScalar().getValue());
+      switch (r.getName()) {
+        case "cpus":
+          cpus = String.valueOf(r.getScalar().getValue());
           break;
-        case RANGES:
-          value = rangesToString(r.getRanges());
+        case "mem":
```
erikdw: Definitely wrong.  The implementation in storm-mesos is better than what you are copying into here.  The existing implementation is more generic.  Please revert.

dsKarthick:
```java
String s = "test"
```

dsKarthick:
@jessicalhartog @erikdw @brndnmtthws @DarinJ Please take a look at the updated changes when you get a chance

storm/src/main/storm/mesos/schedulers/OfferResources.java
erikdw: @dsKarthick : did you end up adding a unit test for `OfferResources` to handle the cases that `testGetResources()` was checking for?

dsKarthick: @erikdw Yup. https://github.com/mesos/storm/pull/93/files#diff-feae3590035962bb8134567db2de626fR38

erikdw: thanks @dsKarthick

erikdw:
@dsKarthick : LGTM, we'll have some further refactoring with the other change we worked on together.  Maybe wait for @DarinJ or @brndnmtthws for the go ahead on committing.

storm/src/main/storm/mesos/schedulers/DefaultScheduler.java
```diff
@@ -0,0 +1,249 @@
...
+
+      double requestedWorkerCpu = MesosCommon.topologyWorkerCpu(mesosStormConf, topologyDetails);
+      double requestedWorkerMem = MesosCommon.topologyWorkerMem(mesosStormConf, topologyDetails);
+
+      log.info("Trying to find " + slotsNeeded + " slots for " + topologyDetails.getId());
+      if (slotsNeeded == 0) {
+        continue;
+      }
+
+      Map<String, Boolean> supervisorExistsPerNodeMap = new HashMap<>();
```
JessicaLHartog: Could this instead be `Set<String> nodesWithExistingSupervisors`? Then we can save some space by [only adding nodes that have supervisors](https://github.com/dsKarthick/storm/blob/scheduler_framework/storm/src/main/storm/mesos/schedulers/DefaultScheduler.java#L119) to the Set, instead of keeping information for all of the nodes. 

To modify the rest of this to work with the Set instead of the map, we can [just check on existence of a node in the set](https://github.com/dsKarthick/storm/blob/scheduler_framework/storm/src/main/storm/mesos/schedulers/DefaultScheduler.java#L126) and put [only the currentNode into the set](https://github.com/dsKarthick/storm/blob/scheduler_framework/storm/src/main/storm/mesos/schedulers/DefaultScheduler.java#L137).

erikdw: This code has already been totally rewritten in our internal paired-on change.... I wouldn't spend *too* much time on reviewing this part of it, since you're gonna be rewriting/pushing our change for it very soon. ;-)

dsKarthick: @JessicaLHartog I have updated the code with your suggested changes. Lets save further improvements on this function for a separate commit like we discussed earlier.

erikdw:
@dsKarthick : let's just merge this when you're back tomorrow.

dsKarthick:
Squashed commits & merged them. Please see https://github.com/mesos/storm/commit/cdd7264f97457a9150df4a46591f2a5b691dfc0f

PR #94, opened 2016-02-22
create unique task ID per task launch

erikdw:
This is needed to avoid a problem with mesos-slave recovery resulting in LOST tasks.

i.e., we discovered that if you relaunch a topology's task onto the same worker slot (so there are 2 different instances with the same "task ID" that have run), then when the mesos-slave process is recovering, it terminates the task upon finding a "terminal" update in the recorded state of the task. The terminal state having been recorded the 1st time the task with that task ID stopped.

To solve this we ensure all task IDs are unique, by adding a milisecond-granularity timestamp onto the task IDs.

### DIFF
--- a/storm/src/main/storm/mesos/MesosCommon.java
+++ b/storm/src/main/storm/mesos/MesosCommon.java
@@ -19,12 +19,14 @@ package storm.mesos;
 
 import backtype.storm.scheduler.TopologyDetails;
 import com.google.common.base.Optional;
-import org.apache.log4j.Logger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import java.util.*;
+import java.util.HashMap;
+import java.util.Map;
 
 public class MesosCommon {
-  public static final Logger LOG = Logger.getLogger(MesosCommon.class);
+  public static final Logger LOG = LoggerFactory.getLogger(MesosCommon.class);
 
   public static final String WORKER_CPU_CONF = "topology.mesos.worker.cpu";
   public static final String WORKER_MEM_CONF = "topology.mesos.worker.mem.mb";
@@ -51,7 +53,7 @@ public class MesosCommon {
   public static String hostFromAssignmentId(String assignmentId, String delimiter) {
     final int last = assignmentId.lastIndexOf(delimiter);
     String host = assignmentId.substring(last + delimiter.length());
-    LOG.debug("assignmentId=" + assignmentId + " host=" + host);
+    LOG.debug("assignmentId={} host={}", assignmentId, host);
     return host;
   }
 
--- a/storm/src/main/storm/mesos/MesosCommon.java
+++ b/storm/src/main/storm/mesos/MesosCommon.java
@@ -74,8 +74,15 @@ public class MesosCommon {
         .or(DEFAULT_MESOS_COMPONENT_NAME_DELIMITER);
   }
 
+  public static String timestampMillis() {
+    long now = System.currentTimeMillis();
+    long sec = now / 1000L;
+    long msec = now % 1000L;
+    return String.valueOf(sec) + "." + String.valueOf(msec);
+  }
+
   public static String taskId(String nodeid, int port) {
-    return nodeid + "-" + port;
+    return nodeid + "-" + port + "-" + timestampMillis();
   }
 
   public static String supervisorId(String nodeid, String topologyId) {
@@ -87,9 +94,30 @@ public class MesosCommon {
   }
 
   public static int portFromTaskId(String taskId) {
-    int last = taskId.lastIndexOf("-");
-    String port = taskId.substring(last + 1);
-    return Integer.parseInt(port);
+    String[] parts = taskId.trim().split("-");
+    if (parts.length < 3) {
+      throw new IllegalArgumentException(String.format("TaskID %s is invalid. " +
+          "Number of dash-delimited components (%d) is less than expected. " +
+          "Expected format is HOSTNAME-PORT-TIMESTAMP", taskId.trim(), parts.length));
+    }
+
+    // TaskID format: HOSTNAME-PORT-TIMESTAMP. Notably, HOSTNAME can have dashes too,
+    // so the port is the 2nd-to-last part after splitting on dash.
+    String portString = parts[parts.length - 2];
+    int port;
+    try {
+      port = Integer.parseInt(portString);
+    } catch (NumberFormatException e) {
+      LOG.error(String.format("Failed to parse string (%s) that was supposed to contain a port.",
+          portString));
+      throw e;
+    }
+
+    if (port < 0 || port > 0xFFFF) {
+      throw new IllegalArgumentException(String.format("%d is not a valid port number.", port));
+    }
+
+    return port;
   }
 
   public static int getSuicideTimeout(Map conf) {
--- a/storm/src/main/storm/mesos/MesosCommon.java
+++ b/storm/src/main/storm/mesos/MesosCommon.java
@@ -24,6 +24,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 public class MesosCommon {
   public static final Logger LOG = LoggerFactory.getLogger(MesosCommon.class);
@@ -76,9 +77,9 @@ public class MesosCommon {
 
   public static String timestampMillis() {
     long now = System.currentTimeMillis();
-    long sec = now / 1000L;
-    long msec = now % 1000L;
-    return String.valueOf(sec) + "." + String.valueOf(msec);
+    long secs = TimeUnit.MILLISECONDS.toSeconds(now);
+    long msecs = now - TimeUnit.SECONDS.toMillis(secs);
+    return String.format("%d.%03d", secs, msecs);
   }
 
   public static String taskId(String nodeid, int port) {
--- a/storm/src/main/storm/mesos/MesosSupervisor.java
+++ b/storm/src/main/storm/mesos/MesosSupervisor.java
@@ -17,7 +17,6 @@
  */
 package storm.mesos;
 
-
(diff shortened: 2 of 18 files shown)


erikdw:
@dsKarthick : please :eyes:

storm/src/main/storm/mesos/MesosSupervisor.java
```diff
@@ -167,8 +189,26 @@ public void launchTask(ExecutorDriver driver, TaskInfo task) {
...
+        port = MesosCommon.portFromTaskId(id.getValue());
+      } catch (IllegalArgumentException e) {
+        LOG.error("killTask: Halting executor process because we had a problem" +
+            " extracting the port from the TaskID: " + id.getValue(), e);
+        Runtime.getRuntime().halt(1);
+      }
+
+      LOG.info("killTask: killing task " + id.getValue() +
+          " which is running on port " + port);
+      _state.remove(Integer.toString(port));
```
erikdw: Unfortunately, this breaks `killedWorker()`.  The intent is to allow MesosSupervisor to communicate to storm-core supervisor that it no longer has this port assigned, and thus should kill the associated worker process.  Unfortunately, by removing the port from `_state`, we prevent `killedWorker()` from issuing the status update later (which needs the TaskID).  So I'm going to change from using `LocalState _state` for this [1] to just using a memory-based map pointing to a tuple <TaskID, Boolean> (boolean indicating whether the port is currently assigned here).  Then I'll change the methods that interact with `_state` to deal with that tuple.

[1] LocalState stores persistently into disk, which doesn't have any value for the MesosSupervisor, since it cannot be restarted.

storm/src/main/storm/mesos/MesosCommon.java
```diff
@@ -72,8 +72,15 @@ public static String getMesosComponentNameDelimiter(Map conf, TopologyDetails in
         .or(DEFAULT_MESOS_COMPONENT_NAME_DELIMITER);
   }
 
+  public static String timestampMillis() {
+    long now = System.currentTimeMillis();
+    long sec = now / 1000L;
+    long msec = now % 1000L;
```
dsKarthick: [This](http://stackoverflow.com/questions/625433/how-to-convert-milliseconds-to-x-mins-x-seconds-in-java) approach seems cleaner?

storm/src/main/storm/mesos/MesosCommon.java
```diff
@@ -72,8 +72,15 @@ public static String getMesosComponentNameDelimiter(Map conf, TopologyDetails in
...
+    return String.valueOf(sec) + "." + String.valueOf(msec);
+  }
+
   public static String taskId(String nodeid, int port) {
-    return nodeid + "-" + port;
+    return nodeid + "-" + port + "-" + timestampMillis();
```
dsKarthick: `String.format`?

storm/src/main/storm/mesos/MesosCommon.java
```diff
@@ -72,8 +72,15 @@ public static String getMesosComponentNameDelimiter(Map conf, TopologyDetails in
         .or(DEFAULT_MESOS_COMPONENT_NAME_DELIMITER);
   }
 
+  public static String timestampMillis() {
```
dsKarthick: How about this

```
 String now = String.valueOf(System.currentTimeMillis());
 String s = String.format("%s-%s", now.substring(0, now.length()-3), now.substring(now.length()-3, now.length()));
```

storm/src/main/storm/mesos/MesosCommon.java
```diff
@@ -85,9 +92,29 @@ public static boolean startLogViewer(Map conf) {
...
 
   public static int portFromTaskId(String taskId) {
-    int last = taskId.lastIndexOf("-");
-    String port = taskId.substring(last + 1);
-    return Integer.parseInt(port);
+    String[] parts = taskId.trim().split("-");
+    if (parts.length < 3) {
+      throw new IllegalArgumentException("TaskID " + taskId.trim() + " is invalid. " +
+          "Number of dash-delimited components (" + parts.length + ") is less than expected. " +
+          "Expected format is HOSTNAME-PORT-TIMESTAMP");
```
dsKarthick: Should we return here? How do we deal with backwards compatibility?

erikdw: There is no point in attempting to do that.  It's a brand new version of the framework, you *must* either launch a new framework or kill all existing executors.

storm/src/main/storm/mesos/MesosCommon.java
erikdw: Let's file an issue to improve the String usage everywhere in this project. 
I don't want more scope creep in this commit (I'm working on adding support for `killTask()` & `shutdown()` to actually do stuff, maybe I should just keep those separate for now).
We should also switch to slf4j-api logging style, instead of the horrendous a + b + c stuff.

storm/src/main/storm/mesos/MesosCommon.java
erikdw: I feel this is a bit _more_ obfuscated than the existing logic.  At the minimum I would break them out like this:

```
String now = String.valueOf(System.currentTimeMillis());
String secs = now.substring(0, now.length()-3);
String msecs = now.substring(now.length()-3, now.length());
return String.format("%s.%s", secs, msecs);
```

Is that better than the / and %?  Maybe it is?  `¯\_(ツ)_/¯`

storm/src/main/storm/mesos/MesosCommon.java
```diff
@@ -85,9 +92,29 @@ public static boolean startLogViewer(Map conf) {
...
+    }
+
+    // TaskID format: HOSTNAME-PORT-TIMESTAMP. Notably, HOSTNAME can have dashes too,
+    // so the port is the 2nd-to-last part after splitting on dash.
+    String portString = parts[parts.length - 2];
```
dsKarthick: Could do `parts[1]` because thats where we expect the port to be

storm/src/main/storm/mesos/MesosSupervisor.java
```diff
@@ -107,7 +110,7 @@ public Object getMetadata() {
     for (int i = 0; i < ports.length; i++) {
       p[i] = Integer.parseInt((String) ports[i]);
     }
-    return PersistentVector.create(p);
+    return PersistentVector.create((Object[]) p);
```
dsKarthick: Why do you explicitly downcast p?

storm/src/main/storm/mesos/MesosSupervisor.java
```diff
@@ -122,10 +125,12 @@ public String getAssignmentId() {
 
   @Override
   public void killedWorker(int port) {
+    LOG.info("killedWorker: removing port " + port + " from the 'assigned port state'");
+    String taskId = _state.get(Integer.toString(port));
```
dsKarthick: `Integer.toString(port)` is kind of a non standard way of using `toString` method. I think `String.valueOf(port)` would be better.

storm/src/main/storm/mesos/MesosCommon.java
dsKarthick: Yes.

storm/src/main/storm/mesos/MesosCommon.java
JessicaLHartog: The comment addresses this. 
```
// TaskID format: HOSTNAME-PORT-TIMESTAMP. Notably, HOSTNAME can have dashes too,
// so the port is the 2nd-to-last part after splitting on dash.
```

dsKarthick: @JessicaLHartog Not sure what you mean - I was just proposing not to do a subtraction.

JessicaLHartog: What I means is that we need to do a subtraction. 
If HOSTNAME has a hyphen, say `my-hyphenated-hostname`, then the taskID would be `my-hyphenated-hostname-PORT-TIMESTAMP` and `parts[1]` would be `hyphenated` and not `PORT`

dsKarthick: @JessicaLHartog Huh.. Good point.

storm/src/main/storm/mesos/MesosSupervisor.java
erikdw: There was a warning that told me to do it.  I believed it. ;-P

storm/src/main/storm/mesos/MesosSupervisor.java
erikdw: I'm just mimicking the existing style, but yes, I agree that String.valueOf() is more canonical.  Also, this is going away with the rewrite to avoid using LocalState.  Please review our internal change first.  As expressed offline, I thought this review would be trivial, hence why I put this public one up already.

storm/src/main/storm/mesos/MesosSupervisor.java
erikdw: Also, just for accuracy, this is "upcasting", not "downcasting".  Downcasting means converting to a more specific derived type:
* https://en.wikipedia.org/wiki/Downcasting

storm/src/main/storm/mesos/MesosCommon.java
erikdw: * slf4j-api: https://github.com/mesos/storm/issues/98
* String.format: https://github.com/mesos/storm/issues/99

Also, thanks for the recommendation on TimeUnit, I've switched to this instead of the math-ey logic:

```
    long now = System.currentTimeMillis();
    long secs = TimeUnit.MILLISECONDS.toSeconds(now);
    long msecs = (TimeUnit.MILLISECONDS.toMillis(now) -
                            TimeUnit.SECONDS.toMillis(TimeUnit.MILLISECONDS.toSeconds(now)));
    return String.format("%d.%03d", secs, msecs);
```

erikdw:
This is being continued in #106

Issue #102, opened 2016-03-08
Logs from mesos/storm repo are lost


dsKarthick:
When I was playing around with vagrant, I noticed that we are losing the logs generated by mesos/storm repo. That is, the mesos/storm appication logs do not appear in the `nimbus.log` but logs from storm-core still appear in the `nimbus.log`.

dsKarthick:
The bug turns out to be related to shading. When we shade `org.slf4j`, we end up shading all the classes belonging to `slf4j-api` provided by storm-core. This causes slf4j not to bind to concrete implementation provided by storm-core. Eg. org.slf4j.Logger interface in slf4j-4j becomes org.slf4j.shaded.Logger and none of the Logger in the storm jar would implement org.slf4j.shaded.Logger.

erikdw:
#102 was fixed by #103.

PR #103, opened 2016-03-08
Bug fix - Prevent losing logs due to incorrect shading

dsKarthick:
As mentioned in https://github.com/mesos/storm/issues/102, when we shade org.slf4j, we end up shading all the classes belonging to slf4j-api provided by storm-core. This causes slf4j not to bind to concrete implementation provided by storm-core. Eg. org.slf4j.Logger interface in slf4j-4j becomes org.slf4j.shaded.Logger and none of the Logger in the storm jar would implement org.slf4j.shaded.Logger. Hence all the log operations in the storm/mesos translates to NOP's. This PR aims to fix it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -145,6 +145,18 @@
         <groupId>org.slf4j</groupId>
         <artifactId>log4j-over-slf4j</artifactId>
         <version>1.6.6</version>
+        <exclusions>
+        <!-- slf4j-api is provided by storm-core. We might as well
+               use the one that is provided to us by storm rather than including
+               the slf4j-api classes in the jar.
+               WARNING: Please do *not* shade org.slf4j. Classes from org.slf4j
+                        are referenced within log4j-over-slf4j and shading them
+                        would lead to java.lang.NoClassDefFoundError -->
+          <exclusion>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+          </exclusion>
+        </exclusions>
       </dependency>
       <dependency>
         <groupId>org.mortbay.jetty</groupId>
--- a/storm/pom.xml
+++ b/storm/pom.xml
@@ -70,6 +70,19 @@
             </filter>
           </filters>
           <relocations>
+          <!-- List of dependencies that should not be shaded
+                 1. org.slf4j:
+                     shading org.slf4j would lead to a couple of issues:
+                     a) Shading org.slf4j corresponding to slf4j-api would prevent its classes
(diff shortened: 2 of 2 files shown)


dsKarthick:
@erikdw @brndnmtthws please :eyes:

brndnmtthws:
Makes sense. Can you describe how you tested it?

dsKarthick:
@brndnmtthws I tested it using the vagrant cluster

brndnmtthws:
Cool, :shipit:.

pom.xml
```diff
@@ -145,6 +145,18 @@
         <groupId>org.slf4j</groupId>
         <artifactId>log4j-over-slf4j</artifactId>
         <version>1.6.6</version>
+	<exclusions>
+          <!-- slf4j-api is provided by storm-core. Including slf4j-api
+	    here would cause its classes to be shaded preventing itself
```
erikdw: I feel like the description in this comment slightly conflicts with the change in `storm/pom.xml` below (which removes the shading of `org.slf4j`).  i.e., do we need both of these changes?

dsKarthick: Ah. Good point. I dont think we need the change to `storm/pom.xml`. Let me revert  `storm/pom.xml` changes and make sure it works.

dsKarthick: @erikdw Ok. So reverting `storm/pom.xml` cause MesosNimbus to throw exception like `java.lang.NoClassDefFoundError: org/slf4j/shaded/MarkerFactory` because of references like ` org.slf4j.MarkerFactory` in [here](http://grepcode.com/file/repo1.maven.org/maven2/org.slf4j/log4j-over-slf4j/1.6.6/org/apache/log4j/Category.java?av=f). So lets keep this change. I will add my reason for excluding slf4j-api in the description.

storm/pom.xml
```diff
@@ -69,7 +69,20 @@
               </excludes>
             </filter>
           </filters>
-          <relocations>
+	  <relocations>
```
erikdw: This is shifted 1 column left, right?

pom.xml
```diff
@@ -145,6 +145,18 @@
         <groupId>org.slf4j</groupId>
         <artifactId>log4j-over-slf4j</artifactId>
         <version>1.6.6</version>
+	<exclusions>
```
erikdw: This appears to not be aligned correctly.

dsKarthick:
@erikdw Fixed.

erikdw:
Thanks @dsKarthick !  LGTM, please merge at your convenience.

PR #106, opened 2016-03-16
Unique TaskIDs

erikdw:
## create unique task ID per task launch
This is needed to avoid a problem with mesos-slave recovery resulting in LOST tasks.

i.e., we discovered that if you relaunch a topology's task onto the same worker slot (so there are 2 different instances with the same "task ID" that have run), then when the mesos-slave process is recovering, it terminates the task upon finding a "terminal" update in the recorded state of the task. The terminal state having been recorded the 1st time the task with that task ID stopped.

To solve this we ensure all task IDs are unique, by adding a milisecond-granularity timestamp onto the task IDs.

### DIFF
--- a/storm/src/main/storm/mesos/MesosSupervisor.java
+++ b/storm/src/main/storm/mesos/MesosSupervisor.java
@@ -17,7 +17,6 @@
  */
 package storm.mesos;
 
-
 import backtype.storm.scheduler.ISupervisor;
 import backtype.storm.utils.Utils;
 import clojure.lang.PersistentVector;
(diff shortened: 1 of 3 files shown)


erikdw:
continuation of #94 : had problems merging and rebasing and gave up. :)  Just cherry-picked my changes into this new branch.  @dsKarthick & @JessicaLHartog: please :eyes: at your convenience.

dsKarthick:
@jessicaLHartog

dsKarthick:
:shipit:

PR #109, opened 2016-03-21
onbuild triggers for mesos/storm versioning and building

lfundaro:
By deferring the building of mesos/storm one is able to use mesos/storm image as a base image and then being able to easily control the desired versioning. 
Instead of having to pull the mesos/storm repo and build the container passing the `--build-arg STORM_RELEASE/MESOS_RELEASE` I find more useful to just use an onbuild version of Dockerfile that allow me to simply do on my own dockerfile the following: 

```
FROM mesos/storm-onbuild 
....
<here I added my private things>
.....
```
and then
`docker build -t my_storm_mesos_image --build-arg STORM_RELEASE=0.10.0 .
`
what do you guys think ?

### DIFF
--- a/onbuild/Dockerfile
+++ b/onbuild/Dockerfile
@@ -1,0 +1,29 @@
+#
+# Dockerfile for Storm Mesos framework
+#
+FROM mesosphere/mesos:0.27.0-0.2.190.ubuntu1404
+MAINTAINER Timothy Chen <tnachen@gmail.com>
+
+ENV DEBIAN_FRONTEND noninteractive
+
+# export environment
+ENV JAVA_HOME /usr/lib/jvm/java-7-openjdk-amd64
+
+ENV MESOS_NATIVE_JAVA_LIBRARY /usr/lib/libmesos.so
+
+ONBUILD ARG MESOS_RELEASE=0.27.0
+ONBUILD ARG STORM_RELEASE=0.9.6
+ONBUILD ARG MIRROR=http://www.gtlib.gatech.edu/pub
+
+ADD . /tmp
+
+RUN apt-get update && \
+  apt-get install -y openjdk-7-jdk maven wget
+ONBUILD RUN cd /tmp && ./bin/build-release.sh main && \
+            mkdir -p /opt/storm && \
+            tar xf /tmp/storm-mesos-*.tgz -C /opt/storm --strip=1 && \
+            rm -rf /tmp/* ~/.m2 && \
+            apt-get -yf autoremove openjdk-7-jdk maven && \
+            apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+WORKDIR /opt/storm
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,21 +11,19 @@ ENV JAVA_HOME /usr/lib/jvm/java-7-openjdk-amd64
 
 ENV MESOS_NATIVE_JAVA_LIBRARY /usr/lib/libmesos.so
 
-ARG MESOS_RELEASE=0.27.0
-ARG STORM_RELEASE=0.9.6
-ARG MIRROR=http://www.gtlib.gatech.edu/pub
+ONBUILD ARG MESOS_RELEASE=0.27.0
+ONBUILD ARG STORM_RELEASE=0.9.6
+ONBUILD ARG MIRROR=http://www.gtlib.gatech.edu/pub
 
-ADD . /work
-
-WORKDIR /work
+ADD . /tmp
 
 RUN apt-get update && \
-  apt-get install -y openjdk-7-jdk maven wget && \
-  ./bin/build-release.sh main && \
-  mkdir -p /opt/storm && \
-  tar xf storm-mesos-*.tgz -C /opt/storm --strip=1 && \
-  rm -rf /work ~/.m2 && \
-  apt-get -yf autoremove openjdk-7-jdk maven && \
-  apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+  apt-get install -y openjdk-7-jdk maven wget
+ONBUILD RUN cd /tmp && ./bin/build-release.sh main && \
+            mkdir -p /opt/storm && \
+            tar xf /tmp/storm-mesos-*.tgz -C /opt/storm --strip=1 && \
+            rm -rf /tmp/* ~/.m2 && \
+            apt-get -yf autoremove openjdk-7-jdk maven && \
+            apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 WORKDIR /opt/storm
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,19 +11,21 @@ ENV JAVA_HOME /usr/lib/jvm/java-7-openjdk-amd64
 
 ENV MESOS_NATIVE_JAVA_LIBRARY /usr/lib/libmesos.so
 
-ONBUILD ARG MESOS_RELEASE=0.27.0
-ONBUILD ARG STORM_RELEASE=0.9.6
-ONBUILD ARG MIRROR=http://www.gtlib.gatech.edu/pub
+ARG MESOS_RELEASE=0.27.0
+ARG STORM_RELEASE=0.9.6
+ARG MIRROR=http://www.gtlib.gatech.edu/pub
 
-ADD . /tmp
+ADD . /work
+
+WORKDIR /work
 
 RUN apt-get update && \
-  apt-get install -y openjdk-7-jdk maven wget
-ONBUILD RUN cd /tmp && ./bin/build-release.sh main && \
-            mkdir -p /opt/storm && \
-            tar xf /tmp/storm-mesos-*.tgz -C /opt/storm --strip=1 && \
-            rm -rf /tmp/* ~/.m2 && \
-            apt-get -yf autoremove openjdk-7-jdk maven && \
-            apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+  apt-get install -y openjdk-7-jdk maven wget && \
+  ./bin/build-release.sh main && \
+  mkdir -p /opt/storm && \
+  tar xf storm-mesos-*.tgz -C /opt/storm --strip=1 && \
+  rm -rf /work ~/.m2 && \
+  apt-get -yf autoremove openjdk-7-jdk maven && \
+  apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 WORKDIR /opt/storm


DarinJ:
Looking at it briefly I think it will work for storm versions but not mesos versions as we're still relying on /usr/local/lib/libmesos.so from the docker image.  Am I missing something?

lfundaro:
@DarinJ  do you mean that it could break in the long run because of the hardcoding on the libmesos.so library ? If this is the case, yes, and it could happen already on the current container because we're making 
``ARG MESOS_RELEASE=0.27.0`` which leaves an open door for someone to pass a different version through  --build-arg when building the image. 
We could then leave the MESOS_RELEASE strictly as an ENV var and the STORM_RELEASE as ARG so people can change it.

erikdw:
@lfundaro : this is an interesting proposal for sure!  Seems like using `ONBUILD` would be useful, but I wanna make sure that I understand the motivation correctly, and that we do this onbuild-ification of the Dockerfile in a canonical fashion.  Towards that end:

1. Can you please give me an example of what kind of "private things" you'd be doing in this downstream Dockerfile?  I have a guess that it might be stuff like modifying the `storm.yaml` file's contents?
2. It seems from [Dockerfile best practices - ONBUILD](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#onbuild) that we should have a separate docker image with an "-onbuild" suffix of the image name, and that we might wanna also have a separate Dockerfile for the onbuild use-case.

lfundaro:
@erikdw 
1. By "private things" I refer as you guessed to storm.yaml settings. But I can imagine that someone might want to put a newrelic agent ? Or whoKnowsWhatMore ? That's why I might find it useful to have an onbuild dockerfile. 
2. Yes I agree, let me propose you a change following this conventions for "onbuild" and let's keep iterating over it. 

I'll ping you when I have it.

lfundaro:
@erikdw So I reverted the changes on the original Dockerfile and added a folder for an onbuild Dockerfile. From what you can see in the [golang docker repository](https://github.com/docker-library/golang) they use several folders for the version of the language and within those folders they have a standard Dockerfile and an onbuild folder with the corresponding onbuild Dockerfile. In our case I don't think we can make this distinction unless we turn ``MESOS_RELEASE`` into something you cannot change, e.g.:

```
mesos-0.26
         | __ onbuild
         |___Dockerfile

  mesos-0.27
         |___onbuild
         |___Dockerfile
```
and so on. I don't know if this makes sense or should we leave the possibility to the user to choose whatever combination of `STORM_RELEASE` and `MESOS_RELEASE`  want ?

onbuild/Dockerfile
```diff
@@ -0,0 +1,29 @@
+#
+# Dockerfile for Storm Mesos framework
+#
+FROM mesosphere/mesos:0.27.0-0.2.190.ubuntu1404
```
lfundaro: I don't understand why we stick to a base image that has a sticked version of mesos and the we let the user choose the ``MESOS_RELEASE`` and ``STORM_RELEASE``

DarinJ: I think we considered using a stock base image and installing mesos as part of the docker build.

erikdw: Yup, we do *want* to do that!   Unfortunately, the mesos packages published by Mesosphere do not have a predictable name for each version (note the "-0.2.190" portion above, it is present in the published rpm/debs as well, and is an artifact from the build process used by Mesosphere).  I've raised that issue with Mesosphere already and they are supposedly planning on getting rid of these extra obfuscating bits.

erikdw: So an implication is that for now, we'd need some sort of forking/switch logic to handle the varying names.  Not sure how that fits in with the workflow of building a Docker image.  e.g., I could envision adding logic to  `bin/build-release.sh` that populates the package name based on lookup table, and then passing that as a variable into the Docker build.

erikdw: Ah, @lfundaro, I lazily responded before reading your [earlier message of the day](https://github.com/mesos/storm/pull/109#issuecomment-200314973").  So, having varying dirs for Dockerfiles for each mesos version is another potential solution/workaround for handling mesos version differences, and probably more "canonical" Docker-workflow-wise.  What say you @DarinJ & @dsKarthick & @brndnmtthws?

DarinJ: That sounds like a pretty good solution to me.  We'd have to worry about updating the hierarchy tree during mesos releases, but it not a difficult task.

DarinJ: I will propose a second solution as well after playing with apt-get a little.  We could use a stock java/ubuntu image and install the mesosphere repo as described [here](https://mesosphere.com/downloads/). Then install the proper version of mesos with:
```
apt-get install mesos=`apt-cache madison mesos | grep ${MESOS_RELEASE} | head -1 | cut -d'|' -f2 | cut -d' ' -f2`
```
`apt-cache madison mesos` displays all the versions of mesos i.e.
```
     mesos | 0.28.0-2.0.16.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
     mesos | 0.27.2-2.0.15.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
     mesos | 0.27.1-2.0.226.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
     mesos | 0.27.0-0.2.190.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
     mesos | 0.26.0-0.2.145.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
     mesos | 0.25.0-0.2.70.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
     mesos | 0.24.1-0.2.35.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
     mesos | 0.24.0-1.0.27.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
     mesos | 0.23.1-0.2.61.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
     mesos | 0.23.0-1.0.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
     mesos | 0.22.2-0.2.62.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
     mesos | 0.22.1-1.0.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
     mesos | 0.22.0-1.0.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
     mesos | 0.21.2-0.2.63.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
     mesos | 0.21.1-1.1.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
     mesos | 0.21.0-1.0.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
     mesos | 0.20.1-1.0.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
     mesos | 0.20.0-1.0.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
     mesos | 0.19.1-1.0.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
     mesos | 0.19.0-1.1.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
     mesos | 0.19.0-1.0.ubuntu1404 | http://repos.mesosphere.io/ubuntu/ trusty/main amd64 Packages
```

erikdw: Whoa, that's a fancy / nice idea!  I think that would work pretty well.  Thanks for the idea & quick hacking!  I'd be happy to accept a PR for this or do it myself later (a bit busy at moment, per usual).

(BTW, the new Mesosphere site layout looks really nice, *but* it would be great if we could link to the actual thing we've clicked into on [that page](https://mesosphere.com/downloads/#) -- i.e., "Get Started" under Mesos.)

dsKarthick: @DarinJ Awesome! Exactly what we have been wanting!

@erikdw @lfundaro @DarinJ With @DarinJ's finding, we probably dont even need a separate folder for versioning - do we? If so, I am slightly concerned about having a folder for mesos version but not for storm version. Thoughts?

DarinJ: @dsKarthick Yeah, I think this would eliminate the need for any separate folders, build-release.sh can just pass the MESOS_RELEASE via build-args.  We could actually cut Docker releases with different tags this way.

erikdw: @dsKarthick & @DarinJ : agreed, this will help with a couple of pending issues.

(BTW, @dsKarthick , the reason it *might* have made sense to have the separated folders for Mesos vs. Storm is that we are actually pulling in the Storm bits ourselves in this project's build script -- whereas we rely on Mesos to be in the environment we are running within.  So Mesos is analogous to Go as @lfundaro pointed out.)

lfundaro:
Hey guys, after the discussion I am a bit confused here. I understand @DarinJ 's solution and I think it's a good idea. But does this mean the ONBUILD triggers are no longer needed ? Are we creating separate docker images per mesos version and then let the user choose the storm_release ?

DarinJ:
@lfundaro I think we want the triggers but also a change to FROM ubuntu and use apt to install the appropriate mesos version base off the MESOS_RELEASE build argument. So
```
docker build -t my_storm_mesos_image --build-arg STORM_RELEASE=0.10.0 --build-arg=0.26.1 . 
```
will install mesos-0.26.0 (and the appropriate libmesos.so in the process.

I'd wait for confirmation from @erikdw and @dsKarthick before doing to much work.

erikdw:
My thoughts were aligned with (DarinJ's comment)[https://github.com/mesos/storm/pull/109#issuecomment-200890109].   However, that is for the "directly runnable docker image" Dockerfile (non-`onbuild`).  For the `onbuild` Docker images, we would indeed be deferring the resolution of the STORM_RELEASE to the building of the child Dockerfile.

P.S., Minor typo correction to @DarinJ's `docker build` invocation (adding ` MESOS_RELEASE` before `=0.26.1`):

```
docker build -t my_storm_mesos_image --build-arg STORM_RELEASE=0.10.0 --build-arg MESOS_RELEASE=0.26.1 . 
```

dsKarthick:
@DarinJ @erikdw +1

lfundaro:
Alright then I wait until the other pull request is ready and merged to add
the onbuild part. Or you might as well add everything into a single pull
request. Cheers

dsKarthick:
@lfundaro I am not sure if you have to be blocked by #83 if you are interested in working on this. Its better to keep the changes separate so we avoid one uber PR with a ton of changes. Also, this could be a quiker change that #83(hopefully).

erikdw:
@lfundaro : looks like @salimane has taken up creation of this onbuild Dockerfile in #120.  Would appreciate your eyes on that review too, thanks!

lfundaro:
Ok ! No problem :). I will check issue #120.


Lorenzo Fundaró García

PR #110, opened 2016-03-23
Consolidate offers & use multiple offers to launch workers

dsKarthick:
There are a few problems with the current scheduling strategy:
    * A lot of times, the offers get too fragmented. So even though there are enough resources, because they are fragmented across multiple offers, they are not useful
    * A lot of times, the offer that is available at the time of allSlotsAvailableForScheduling is in the background by the rotating map before the storm-core invokes schedule to actually schedule the offers. This leads to a race-condition where some of the n workers take a long time to launch. 

This PR fixes both above mentioned issues.


dsKarthick:
@jessicalhartog @erikdw @brndnmtthws @darinJ @tnachen Please take a look when you get a chance.

erikdw:
@dsKarthick is this the 2nd try at this, or a separate PR that is dependent on #93 being merged first and then rebasing?  If it's the "same thing" please try to keep only 1 open at a time, otherwise it's confusing.

dsKarthick:
@erikdw No. https://github.com/mesos/storm/pull/93 is different. This PR is an improvement of https://github.com/mesos/storm/pull/93

dsKarthick:
Rebased this pr to the latest master with the scheduler framework. Please refer to https://github.com/mesos/storm/pull/111

PR #111, opened 2016-03-24
Consolidate offers from the same host into OfferResources. 

dsKarthick:
There are a few problems with the current scheduling strategy:
* A lot of times, the offers get too fragmented. So even though there are enough resources, because they are fragmented across multiple offers, they are not useful
* A lot of times, the offer that is available at the time of allSlotsAvailableForScheduling is in the background by the rotating map before the storm-core invokes schedule to actually schedule the offers. This leads to a race-condition where some of the n workers take a long time to launch.

This PR aims to fix both above mentioned issues.


dsKarthick:
@jessicalhartog @erikdw @brndnmtthws @darinJ @tnachen Please take a look when you get a chance. 

Please note that I have removed some of the tests - I will restore them in next day or so. As a first step, your feedback on high level design would be helpful - especially the design related to the reserved resources.

PR #115, opened 2016-03-25
per-topology supervisor logs

erikdw:
Since storm-mesos launches separate supervisors for each topology on a host,
we should be separating out the supervisor logs on a per-topology basis.

Towards this end, I have leveraged the [STORM-1056](https://issues.apache.org/jira/browse/STORM-1056) change I introduced into the storm project.  This has been released in storm-0.9.6 and storm-0.10.0.
With this change we can pass an env variable to `bin/storm` (or `bin/storm.py` in storm-0.10.0+)
which sets the name of the supervisor log file:
* `STORM_SUPERVISOR_LOG_FILE`
Notably, the *location* that the logs are written to is unchanged.

Previously the logs were written (by default) to:

```
/tmp/mesos/slaves/SLAVE_UUID/frameworks/FRAMEWORK_UUID/executors/TOPOLOGY_ID/runs/RUN_UUID/storm-mesos-*/logs/supervisor.log
```

Now the logs will be written (by default) to:

```
/tmp/mesos/slaves/SLAVE_UUID/frameworks/FRAMEWORK_UUID/executors/TOPOLOGY_ID/runs/RUN_UUID/storm-mesos-*/logs/TOPOLOGY_ID-supervisor.log
```

If you set `storm.log.dir` in `storm.yaml` to a common place (e.g., for ingesting into splunk from a common dir), then it might look like this:

```
 # in storm.yaml
 storm.log.dir: "/var/tmp/storm/logs"

 # supervisor log file path:
 /var/tmp/storm/logs/TOPOLOGY_ID-supervisor.log
```

### DIFF
--- a/storm-shim-10x/src/main/java/storm/mesos/shims/CommandLineShim.java
+++ b/storm-shim-10x/src/main/java/storm/mesos/shims/CommandLineShim.java
@@ -25,9 +25,13 @@ public class CommandLineShim implements ICommandLineShim {
     this.extraConfig = extraConfig;
   }
 
-  public String getCommandLine() {
-    return "cp storm.yaml storm-mesos*/conf && cd storm-mesos* && python bin/storm.py " +
-        "supervisor storm.mesos.MesosSupervisor" + extraConfig;
+  public String getCommandLine(String topologyId) {
+    return String.format(
+        "export STORM_SUPERVISOR_LOG_FILE=%s-supervisor.log" +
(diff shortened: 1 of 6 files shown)


erikdw:
@dsKarthick & @jessicaLHartog: please :eyes:

dsKarthick:
:shipit:

JessicaLHartog:
LGTM! :ship:

PR #117, opened 2016-03-26
Add automated archive builds and GitHub Releases Uploading

### LINKED ISSUE #83: automate release tagging and building

salimane:
This fixes #83 automate release tagging and building.

I've also included in this PR: 
- testing  against multiple JDKs, Oracle JDK 7 and 8 
- testing using a build matrix of recents versions of Storm and Mesos
- automatically uploading releases tarballs to github like https://github.com/salimane/storm/releases

To release a new version of storm-mesos: 
- create a new release branch, e.g. release_0.1.0, with the right versions in the pom.xml files, e.g. not "0.1.0-SNAPSHOT", but "0.1.0"
- push the branch
- go to  https://github.com/mesos/storm/releases, create a new release, choosing the release branch, e.g.  release_0.1.0 and name it for example "0.1.0"
- and then wait for travis to do its magic

By the way, after merging, don't forget to update the deploy api_key included in the change.

After this is merged, I will work on the automated docker images building.

By the way, I've already made the whole pipeline work including docker images building using travis in my fork. You can see the release packages https://github.com/salimane/storm/releases/tag/0.1.3 and the docker images automatically built and pushed to docker hub here https://hub.docker.com/r/salimane/storm-mesos/tags/ 

cc @erikdw 
Thanks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,4 +2,30 @@ language: java
 sudo: false
 cache:
   directories:
-      - $HOME/.m2/repository
+    - "$HOME/.m2/repository"
+
+install: mvn install -DskipTests=true -Dmaven.javadoc.skip=true
+script:
+  - jdk_switcher use oraclejdk8
+  - mvn test
+  - jdk_switcher use oraclejdk7
+  - mvn test
+  - bin/build-release.sh
+
+matrix:
+  fast_finish: true
+  include:
+    - env: STORM_RELEASE="0.10.0" MESOS_RELEASE="0.28.0"
+    - env: STORM_RELEASE="0.10.0" MESOS_RELEASE="0.27.2"
+    - env: STORM_RELEASE="0.9.6" MESOS_RELEASE="0.28.0"
+    - env: STORM_RELEASE="0.9.6" MESOS_RELEASE="0.27.2"
+
+# Deploy archives to GitHub release tags
+deploy:
+  provider: releases
+  api_key:
+    secure: DMwskItRMoVMpmYt3trBy4WlQN37ZnT28p1jYhP6xQOD1CfgZC+ggGiWr+aobfD89ZvztGdRPXyBCorsErqXEPLIA/dSAfJaOr20o6KzyFz284vw08dLwIvjev/eRYuT5BuRVCyFkt4MkTQEEGJdBE1Fb0nc8xdaylDR6CqthyI/qEE1dx4/yFoyBTBHQOBI3EynOG+h3D6XKQCfheNyX3i63St2wyI0hfTnanGNpsdxwgO5H4SW7p48/f+z4kjIt9IxHL7Lr8b56WbKvo97OqXzn+OFiCO8m/DtDD+HTUqjwDQA5oYvMx9bBHmZEyz/HgTnNsgH0RqTfo97AqezV37LSzJyWupF6Nk5N3k+UJu5ck/EvaiOBjDsye9sM1O6W5XICUfgNKHHvx6Ow5nVBY7gyrcCsJ+w65sda6Ij731GxmWEv2u9rfpq0/mNRvPwf/wFS9JN7SwNzsJPTYswsNyJoH7o/SVH7CiRavUTyCJJvelSimY9GwOWBCVNQIR8cioOncqH1mp1oxbSAVutcoppUt/jDNrxQwFM+XeI8AifK/CwoujSR6e3QKCPyRh3t5QICHwDcAh5qKDNGrLKEjTejCX+MdDKro+qzh4w7/T8jcZEKS+TpPC3cnFiT5N4joJ0EQ7oENRLpTR2JA4K38EPMSluma2dzveTXdz74QE=
+  file: "storm-mesos-$(grep -1 -A 0 -B 0 '<version>' pom.xml | head -n 1 | awk '{print $1}' | sed -e 's/.*<version>//' | sed -e 's|</version>.*||')-storm$STORM_RELEASE-mesos$MESOS_RELEASE.tgz"
+  skip_cleanup: true
+  on:
+    tags: true


.travis.yml
```diff
@@ -2,4 +2,30 @@ language: java
...
+install: mvn install -DskipTests=true -Dmaven.javadoc.skip=true
+script:
+  - jdk_switcher use oraclejdk8
+  - mvn test
+  - jdk_switcher use oraclejdk7
+  - mvn test
+before_deploy: bin/build-release.sh
+
+matrix:
+  fast_finish: true
```
dsKarthick: Matrix is cool

.travis.yml
```diff
@@ -2,4 +2,30 @@ language: java
...
+    - env: STORM_RELEASE="0.10.0" MESOS_RELEASE="0.28.0"
+    - env: STORM_RELEASE="0.10.0" MESOS_RELEASE="0.27.2"
+    - env: STORM_RELEASE="0.9.6" MESOS_RELEASE="0.28.0"
+    - env: STORM_RELEASE="0.9.6" MESOS_RELEASE="0.27.2"
+
+# Deploy archives to Github release tags
+deploy:
+  provider: releases
+  api_key:
+    secure: DMwskItRMoVMpmYt3trBy4WlQN37ZnT28p1jYhP6xQOD1CfgZC+ggGiWr+aobfD89ZvztGdRPXyBCorsErqXEPLIA/dSAfJaOr20o6KzyFz284vw08dLwIvjev/eRYuT5BuRVCyFkt4MkTQEEGJdBE1Fb0nc8xdaylDR6CqthyI/qEE1dx4/yFoyBTBHQOBI3EynOG+h3D6XKQCfheNyX3i63St2wyI0hfTnanGNpsdxwgO5H4SW7p48/f+z4kjIt9IxHL7Lr8b56WbKvo97OqXzn+OFiCO8m/DtDD+HTUqjwDQA5oYvMx9bBHmZEyz/HgTnNsgH0RqTfo97AqezV37LSzJyWupF6Nk5N3k+UJu5ck/EvaiOBjDsye9sM1O6W5XICUfgNKHHvx6Ow5nVBY7gyrcCsJ+w65sda6Ij731GxmWEv2u9rfpq0/mNRvPwf/wFS9JN7SwNzsJPTYswsNyJoH7o/SVH7CiRavUTyCJJvelSimY9GwOWBCVNQIR8cioOncqH1mp1oxbSAVutcoppUt/jDNrxQwFM+XeI8AifK/CwoujSR6e3QKCPyRh3t5QICHwDcAh5qKDNGrLKEjTejCX+MdDKro+qzh4w7/T8jcZEKS+TpPC3cnFiT5N4joJ0EQ7oENRLpTR2JA4K38EPMSluma2dzveTXdz74QE=
```
erikdw: I know you mentioned that we'd need to update the api_key after merging, but I'd like a bit more info on this.  Can you please explain how this key was generated, and whether it seems safe/acceptable to have it open for view/editing by non-members of the group?  I'm guessing we don't wanna have it just sitting in our repo open to the world, since I'm assuming that would let anyone do API requests to the mesos/storm repo and potentially break it.

Perhaps we should leverage [travis's support for encrypting stuff](https://docs.travis-ci.com/user/encryption-keys/)? (Which I know *nothing* about!)

.travis.yml
```diff
@@ -2,4 +2,30 @@ language: java
...
 cache:
   directories:
-      - $HOME/.m2/repository
+    - "$HOME/.m2/repository"
+
+install: mvn install -DskipTests=true -Dmaven.javadoc.skip=true
+script:
+  - jdk_switcher use oraclejdk8
+  - mvn test
+  - jdk_switcher use oraclejdk7
```
dsKarthick: @salimane Do we need to support jdk7?

.travis.yml
salimane: @erikdw this is actually the encrypted version of the api_key :). That's why we can commit it to the github repo, it's the recommended way from travis-ci.

After merging, one can update the deploy block by running : 
```
gem install travis
travis setup -f releases
```
It will ask you for github credentials in order to be able to create tokens. more info here https://docs.travis-ci.com/user/deployment/releases/

.travis.yml
salimane: @dsKarthick, well I don't really need it, but may be some people needs it :) . I'm fine with removing it if we choose to.

.travis.yml
erikdw: Aha!  Well, I certainly need to understand this a bit better then, in order to be comfortable with it. ;-)  Thanks for the explanation and tips, give me a bit of time to grok and test.

.travis.yml
erikdw: @salimane : we were talking about leveraging some Java 8 features, so were talking about dropping support for JDK7 for compiling.  But no specific plan yet, so we can change this test later if we decide to do that.

.travis.yml
```diff
@@ -2,4 +2,30 @@ language: java
 sudo: false
```
erikdw: (Dropping this note here to allow for threading (I dislike that GitHub doesn't allow threaded conversations on the base conversation page.))

Ideally I'd like to integrate this logic with the maven release plugin, to avoid needing to futz with the versions manually and potentially avoid having to push manually to a separate release branch.  Do you see that as being a fairly straightforward extension / follow-up to your work here?

On a related note, I'm unclear on exactly how the uploading of the release is "gated" so that it will only happen when we create a release?  (I'm *guessing* however that this is some implicit behavior from the travis handling of the "deploy provider" being "releases" below.)   Also, we do wanna keep having travis builds for normal pushes, I just wanna make sure this change isn't preventing that.

salimane: Currently, with this PR, there would be a travis build (mvn test) : 
- for every master push
- for every branch pushed
- for every pull request

The uploading of the release package only happens when you explicitly go https://github.com/mesos/storm/releases and create a new release. A new tag would automatically be created. Travis would test that tag (mvn test) and if successful, start the "deploy" process. That is, basically, run ``bin/build_release.sh`` and upload the packages to the specific release page.

.travis.yml
```diff
@@ -2,4 +2,30 @@ language: java
...
-      - $HOME/.m2/repository
+    - "$HOME/.m2/repository"
+
+install: mvn install -DskipTests=true -Dmaven.javadoc.skip=true
+script:
+  - jdk_switcher use oraclejdk8
+  - mvn test
+  - jdk_switcher use oraclejdk7
+  - mvn test
+before_deploy: bin/build-release.sh
```
erikdw: Maybe we should move the running of `bin/build-release.sh` to be the last step of the `script` section?

My thinking:  I'm a bit worried about a scenario where there is a bug introduced in `bin/build-release.sh` by a PR, but we don't catch it during the PR's travis builds since it is *implied* by the travis docs that all of the deploy-related steps won't be run until we merge into master.  (The [docs](https://docs.travis-ci.com/user/deployment#Pull-Requests) actually just say the deployment step won't be done for PR builds, and don't specifically say whether the `before_deploy` step would be run for a PR build.)

.travis.yml
```diff
@@ -2,4 +2,30 @@ language: java
...
+
+matrix:
+  fast_finish: true
+  include:
+    - env: STORM_RELEASE="0.10.0" MESOS_RELEASE="0.28.0"
+    - env: STORM_RELEASE="0.10.0" MESOS_RELEASE="0.27.2"
+    - env: STORM_RELEASE="0.9.6" MESOS_RELEASE="0.28.0"
+    - env: STORM_RELEASE="0.9.6" MESOS_RELEASE="0.27.2"
+
+# Deploy archives to Github release tags
```
erikdw: nit: Github -> GitHub

.travis.yml
erikdw: Thanks for the extra explanation, that helped me!  And we can defer the automagic handling 'til later.

erikdw:
@salimane : looks great, sorry it took me awhile to grok this, I've never actually read any of the travis-ci docs before.  Just the super minor casing nit about GitHub, and the question about whether we should move `bin/build-release.sh` from `before_deploy` into the `script` step.

.travis.yml
erikdw: FYI, I found the description in [this thread](https://github.com/travis-ci/travis-ci/issues/2982) *much* better than the [bare travis docs for encryption](https://docs.travis-ci.com/user/encryption-keys/).

erikdw:
@salimane : no rush, not trying to harass you, just wondering if you missed my 2 minor comments above, given that you've been active on #120 ?  I think we're very close to being able to merge this.

.travis.yml
salimane: updated

.travis.yml
salimane: yep, you are right, I've moved the build_release step into the script block

.travis.yml
erikdw: BTW, a clarification that I'm putting here for posterity:  it *seems* that a new release will be created whenever a new tag is created, since the `.travis.yml` file specifies:  `deploy.on.tags: true`.  i.e., so it seems to me that it's not true that the releases will be created "only" when we create a new release on the https://github.com/mesos/storm/releases page.

erikdw:
@salimane : thanks for addressing the comments, I'm merging now.

.travis.yml
salimane: @erikdw yep, you are right, the releases are created on tags. I' assuming you only create tags on new releases.

PR #118, opened 2016-03-26
logviewer support for configurable storm.log.dir

erikdw:
but fall back to hardcoded mesos-sandbox path.

Notably, `storm.log.dir` is the configuration parameter in storm for choosing the log directory to be used (it's set in the [logback (storm-0.9.x)](https://github.com/apache/storm/blob/v0.9.6/logback/cluster.xml#L21) & [log4j 2 (storm-0.10.0+)](https://github.com/apache/storm/blob/master/log4j2/cluster.xml#L25) config files).

### DIFF
--- a/storm/src/main/storm/mesos/logviewer/LogViewerController.java
+++ b/storm/src/main/storm/mesos/logviewer/LogViewerController.java
@@ -40,7 +40,7 @@ public class LogViewerController {
   public LogViewerController(Map conf) {
     port = Optional.fromNullable((Number) conf.get(Config.LOGVIEWER_PORT)).or(8000).intValue();
     setUrlDetector(new SocketUrlDetection(port));
-    logDir = MesosCommon.getStormLogDir(conf);
+    logDir = MesosCommon.getStormLogDir(conf, System.getenv("MESOS_SANDBOX") + "/logs");
   }
 
   /**
--- a/storm/src/main/storm/mesos/logviewer/LogViewerController.java
+++ b/storm/src/main/storm/mesos/logviewer/LogViewerController.java
@@ -31,13 +31,19 @@ import java.util.Map;
 
 public class LogViewerController {
   private static final Logger LOG = LoggerFactory.getLogger(LogViewerController.class);
+  private static final String STORM_LOG_DIR_CONF = "storm.log.dir";
   protected Process process;
   protected SocketUrlDetection urlDetector;
   protected Integer port;
+  protected String logdir;
 
   public LogViewerController(Map conf) {
     port = Optional.fromNullable((Number) conf.get(Config.LOGVIEWER_PORT)).or(8000).intValue();
     setUrlDetector(new SocketUrlDetection(port));
+    logdir = System.getenv("MESOS_SANDBOX") + "/logs";
+    if (conf.containsKey(STORM_LOG_DIR_CONF)) {
+      logdir = (String) conf.get(STORM_LOG_DIR_CONF);
+    }
   }
 
   /**
@@ -96,7 +102,7 @@ public class LogViewerController {
         Paths.get(System.getProperty("user.dir"), "/bin/storm").toString(),
         "logviewer",
         "-c",
-        "storm.log.dir=" + System.getenv("MESOS_SANDBOX") + "/logs",
+        "storm.log.dir=" + logdir,
         "-c",
         Config.LOGVIEWER_PORT + "=" + port
     );
--- a/storm/src/main/storm/mesos/logviewer/LogViewerController.java
+++ b/storm/src/main/storm/mesos/logviewer/LogViewerController.java
@@ -21,6 +21,7 @@ import backtype.storm.Config;
 import com.google.common.base.Optional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import storm.mesos.util.MesosCommon;
 
 import java.io.File;
 import java.io.IOException;
@@ -31,19 +32,15 @@ import java.util.Map;
 
 public class LogViewerController {
   private static final Logger LOG = LoggerFactory.getLogger(LogViewerController.class);
-  private static final String STORM_LOG_DIR_CONF = "storm.log.dir";
   protected Process process;
   protected SocketUrlDetection urlDetector;
   protected Integer port;
-  protected String logdir;
+  protected String logDir;
 
   public LogViewerController(Map conf) {
     port = Optional.fromNullable((Number) conf.get(Config.LOGVIEWER_PORT)).or(8000).intValue();
     setUrlDetector(new SocketUrlDetection(port));
-    logdir = System.getenv("MESOS_SANDBOX") + "/logs";
-    if (conf.containsKey(STORM_LOG_DIR_CONF)) {
-      logdir = (String) conf.get(STORM_LOG_DIR_CONF);
-    }
+    logDir = MesosCommon.getStormLogDir(conf);
   }
 
   /**
@@ -102,17 +99,17 @@ public class LogViewerController {
         Paths.get(System.getProperty("user.dir"), "/bin/storm").toString(),
         "logviewer",
         "-c",
-        "storm.log.dir=" + logdir,
+        "storm.log.dir=" + logDir,
         "-c",
         Config.LOGVIEWER_PORT + "=" + port
     );
 
     // If anything goes wrong at startup we want to see it.
-    Path logPath = Paths.get(System.getenv("MESOS_SANDBOX"), "/logs");
+    Path logPath = Paths.get(logDir);
     if (!logPath.toFile().exists() && !logPath.toFile().mkdirs()) {
       throw new RuntimeException("Couldn't create log directory");
     }
-    File log = Paths.get(System.getenv("MESOS_SANDBOX"), "/logs/logviewer-startup.log").toFile();
+    File log = Paths.get(logDir, "/logviewer-startup.log").toFile();
     pb.redirectErrorStream(true);
     pb.redirectOutput(Redirect.appendTo(log));
     return pb;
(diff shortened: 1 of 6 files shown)


erikdw:
@dsKarthick @jessicalhartog @DarinJ : please review at your convenience

storm/src/main/storm/mesos/logviewer/LogViewerController.java
```diff
@@ -96,7 +102,7 @@ protected ProcessBuilder createProcessBuilder() {
         Paths.get(System.getProperty("user.dir"), "/bin/storm").toString(),
         "logviewer",
         "-c",
-        "storm.log.dir=" + System.getenv("MESOS_SANDBOX") + "/logs",
+        "storm.log.dir=" + logdir,
```
dsKarthick: @erikdw MESOS_SANDBOX is set by Mesos correct? Wondering if we should override `System.getenv("MESOS_SANDBOX")` with logdir only if `storm.log.dir` is configured?

dsKarthick: Ok. I spoke too soon. I see that you have done exactly what I wanted in the LogViewerController function above :)

storm/src/main/storm/mesos/logviewer/LogViewerController.java
```diff
@@ -31,13 +31,19 @@
 
 public class LogViewerController {
   private static final Logger LOG = LoggerFactory.getLogger(LogViewerController.class);
+  private static final String STORM_LOG_DIR_CONF = "storm.log.dir";
```
dsKarthick: Move this to MesosCommon?

storm/src/main/storm/mesos/logviewer/LogViewerController.java
```diff
@@ -31,13 +31,19 @@
...
   protected Integer port;
+  protected String logdir;
 
   public LogViewerController(Map conf) {
     port = Optional.fromNullable((Number) conf.get(Config.LOGVIEWER_PORT)).or(8000).intValue();
     setUrlDetector(new SocketUrlDetection(port));
+    logdir = System.getenv("MESOS_SANDBOX") + "/logs";
+    if (conf.containsKey(STORM_LOG_DIR_CONF)) {
+      logdir = (String) conf.get(STORM_LOG_DIR_CONF);
+    }
```
dsKarthick: Move lines 43:46 to MesosCommon.getLogDirectory?

storm/src/main/storm/mesos/logviewer/LogViewerController.java
```diff
@@ -31,13 +31,19 @@
...
+  private static final String STORM_LOG_DIR_CONF = "storm.log.dir";
   protected Process process;
   protected SocketUrlDetection urlDetector;
   protected Integer port;
+  protected String logdir;
 
   public LogViewerController(Map conf) {
     port = Optional.fromNullable((Number) conf.get(Config.LOGVIEWER_PORT)).or(8000).intValue();
     setUrlDetector(new SocketUrlDetection(port));
+    logdir = System.getenv("MESOS_SANDBOX") + "/logs";
```
DarinJ: Put line 43 in an else block? Slightly better readability.

storm/src/main/storm/mesos/logviewer/LogViewerController.java
DarinJ: Along these lines do we need to change line 111 to:
```
Path logPath = Paths.get(System.getenv("MESOS_SANDBOX"), "/logs");
```  
Similarly change line 115 to:
```
File log = Paths.get(logDir, "/logviewer-startup.log").toFile();
```

storm/src/main/storm/mesos/logviewer/LogViewerController.java
erikdw: Sure, creating `MesosCommon.getStormLogDir()`

storm/src/main/storm/mesos/logviewer/LogViewerController.java
erikdw: Makes sense, doing that (in `MesosCommon.getStormLogDir()`)

storm/src/main/storm/mesos/logviewer/LogViewerController.java
erikdw: Ah, good catch!  Fixing that.

storm/src/main/storm/mesos/logviewer/LogViewerController.java
erikdw: yup yup! ;)

erikdw:
@DarinJ & @dsKarthick : thank you for the review, I just pushed a change.

NOTE: I also noticed some setting of `storm.log.dir` to the mesos sandbox in the Docker command line shim methods, and in the marathon config file, and in the README:

```
% git grep MESOS_SANDBOX        
README.md:MESOS_SANDBOX path, and running the container. For example:
README.md:$ docker run -i --net=host -e PORT0=10000 -e PORT1=10001 -e MESOS_SANDBOX=/var/log/storm -t mesosphere/storm ./bin/run-with-marathon.sh
bin/run-with-marathon.sh:    ./bin/storm ui -c ui.port=$PORT0 -c nimbus.thrift.port=$PORT1 -c storm.log.dir=$MESOS_SANDBOX/logs ${STORM_UI_OPTS:-}
bin/run-with-marathon.sh:  ./bin/storm-mesos nimbus -c mesos.master.url=zk://${MESOS_MASTER_ZK}:2181/mesos -c storm.zookeeper.servers="[\"${MESOS_MASTER_ZK}\"]" -c nimbus.thrift.port=$PORT1  -c storm.log.dir=$MESOS_SANDBOX/logs  ${STORM_NIMBUS_OPTS:-}
storm-shim-10x/src/main/java/storm/mesos/shims/DockerCommandLineShim.java:        " && /bin/cp $MESOS_SANDBOX/storm.yaml conf " +
storm-shim-10x/src/main/java/storm/mesos/shims/DockerCommandLineShim.java:        "-c storm.log.dir=$MESOS_SANDBOX/logs%s",
storm-shim-9x/src/main/java/storm/mesos/shims/DockerCommandLineShim.java:        " && /bin/cp $MESOS_SANDBOX/storm.yaml conf " +
storm-shim-9x/src/main/java/storm/mesos/shims/DockerCommandLineShim.java:        "-c storm.log.dir=$MESOS_SANDBOX/logs%s",
storm/src/main/storm/mesos/util/MesosCommon.java:      return System.getenv("MESOS_SANDBOX") + "/logs";
```

For now I'm going to ignore this, but it probably needs to be readdressed later.

erikdw:
So I *said* I'm going to ignore the Docker command line references, but now I'm thinking that I'll handle it by changing the `MesosCommon.getStormLogDir()` method to take a `String` which will be the default logDir to return.  And then I'll pass the appropriate default from the Docker shims as well as the `LogViewerController`.

NOTE: as an aside, I wonder why the docker command lines have log.dir set explicitly like this, but the non-docker ones do not.

erikdw:
And that latest plan is hitting a roadblock:  `MesosCommon` is not reference-able from within the storm-shim-9x/storm-shim-10x source code. :_(

DarinJ:
Ah looked a bit at the docker code here.  I think the goal here is to keep the logs from being written to the container.  This has a few advantages.  One it keeps the container size from growing (I've seen that fill /var/lib/docker when people throw logs in there). Two it keeps people from having to exec in to look at logs.  I think this might be desired behavior in all real cases.

I'll look at the shim though to see what changes we could make to allow the majority of the code be visible from the shims.  Likely this will involve refactoring anything not calling the shims to a common sub-project.

erikdw:
@DarinJ 

1. Thanks for the explanation about docker -- I don't have any experience with *really* running docker containers under mesos.  What you said makes sense (I think!).
2. Regarding "replacing LocalState with a Map":  the only thing we use the LocalState mechanism for today (post #106) is simply:
  * storage of the framework ID onto disk, so that when we restart MesosNimbus it can reference that value and appear as the same instance of the framework when MesosNimbus reconnects with Mesos.
  * So perhaps instead of a memory-based Map, we should just write the ID to a file.
     * The file path can just be: the [`localDir`](https://github.com/mesos/storm/blob/b6f232a5edfae4b0be7b747731a46bdd4ac2567f/storm/src/main/storm/mesos/MesosNimbus.java#L169) + `/storm-mesos_framework-id`, or some such ish.
       * NOTE: From my observations, `localDir` is the `cwd` of the MesosNimbus process on startup, followed by `storm-local/nimbus/inimbus/`.  [Here is how storm constructs `localDir`.](https://github.com/apache/storm/blob/v0.9.6/storm-core/src/clj/backtype/storm/config.clj#L161-L163)
3. If we do the above, it's probably ok to rip out all the hard work you did for the shims / modules, since as you said the command line construction is probably relatively to do at runtime somehow (just swapping `storm` for `storm.py` afterall!).

DarinJ:
@erikdw sounds fine, if you want me to look at anything let me know.  A bit swamped right now, but should be able to look things over and provide comments.

erikdw:
@DarinJ : great, I'm also swamped (as always :wink: ), but I'll "at" you on the PR when I send it for a cursory review.

erikdw:
Sooo.... I'm planning to just close this without fixing it.

Reasoning:

1. The auto-launching of logviewer is totally broken in storm-mesos even *with* this fix.  See [this comment in #6](https://github.com/mesos/storm/issues/6#issuecomment-112218318), [this other comment in #6](https://github.com/mesos/storm/issues/6#issuecomment-159144835), and [STORM-1342](https://issues.apache.org/jira/browse/STORM-1342) for more info.
2. The [work proposed above](https://github.com/mesos/storm/pull/118#issuecomment-205149587) is a lot for minimal gain given the previous explanation about logviewer being totally broken anyways.
3. It would be a [PITA](https://www.urbandictionary.com/define.php?term=pita&defid=635466) to upgrade from the storm-mesos version prior to [the proposed fix](https://github.com/mesos/storm/pull/118#issuecomment-205149587) above to the one with the proposed fix.  i.e., if people wanna keep their framework running, then we'd have to tell them how to extract their framework ID out of the storm-local binary blob file and put it into whatever other file we are using.

Issue #119, opened 2016-03-30
ConcurrentHashMap serialization/type issue


drewrobb:
Building against current master I'm getting an exception. 
@erikdw the exception is in mk-supervisor and I think is introduced by https://github.com/mesos/storm/pull/106 from using ConcurrentHashMap. I don't know if java 8 is officially supported, but internally we are using it. What I believe to be the issue is described a bit here: https://gist.github.com/AlainODea/1375759b8720a3f9f094

```
2016-03-29 23:39:03.006 b.s.d.supervisor [ERROR] Error on initialization of server mk-supervisor
java.lang.ClassCastException: java.util.concurrent.ConcurrentHashMap$KeySetView cannot be cast to java.lang.Number
	at clojure.lang.RT.longCast(RT.java:1151) ~[clojure-1.6.0.jar:?]
	at clojure.core$long.invoke(core.clj:3195) ~[clojure-1.6.0.jar:?]
	at clojure.core$map$fn__4245.invoke(core.clj:2557) ~[clojure-1.6.0.jar:?]
	at clojure.lang.LazySeq.sval(LazySeq.java:40) ~[clojure-1.6.0.jar:?]
	at clojure.lang.LazySeq.seq(LazySeq.java:49) ~[clojure-1.6.0.jar:?]
	at clojure.lang.LazySeq.count(LazySeq.java:65) ~[clojure-1.6.0.jar:?]
	at clojure.lang.LazySeq.size(LazySeq.java:167) ~[clojure-1.6.0.jar:?]
	at backtype.storm.generated.SupervisorInfo$SupervisorInfoStandardScheme.write(SupervisorInfo.java:1108) ~[storm-core-0.10.0.jar:0.10.0]
	at backtype.storm.generated.SupervisorInfo$SupervisorInfoStandardScheme.write(SupervisorInfo.java:954) ~[storm-core-0.10.0.jar:0.10.0]
	at backtype.storm.generated.SupervisorInfo.write(SupervisorInfo.java:838) ~[storm-core-0.10.0.jar:0.10.0]
	at org.apache.thrift7.TSerializer.serialize(TSerializer.java:79) ~[storm-core-0.10.0.jar:0.10.0]
	at backtype.storm.serialization.GzipThriftSerializationDelegate.serialize(GzipThriftSerializationDelegate.java:41) ~[storm-core-0.10.0.jar:0.10.0]
[...]

PR #121, opened 2016-04-02
Fix MesosSupervisor.getMetadata

erikdw:
This is intended to fix issue #119.

With the introduction of the `TaskAssignments` refactor for creating unique
task IDs (#106), I introduced a couple of bugs in the implementation of
`MesosSupervisor.getMetadata`:

1. The slot counts in the Storm UI were broken -- the return from
   `getMetadata` was always a single element vector, due to using a
   `Set` instead of a java array previously. This was causing the
   `PersistentVector.create(Object ... object)` method to be matched,
   which just puts the passed objects into a vector without iterating
   over their constituent elements.  Since we are passing a single
   `Set` object, we are getting a single element in the resultant vector. 
   So the fix is to just create a `List` and pass that to
   `PersistentVector.create()`.
2. The returned `Object` must be serializable.  Depending on the
   build and runtime environment, the serialization done by the storm
   supervisor during initialization will fail, crashing the supervisor.
   That was happening because we were passing back the
   `ConcurrentHashMap$KeySetView` object, which is not serializable.
   Here too, the fix is to just create a List and pass that to
   `PersistentVector.create()`.

NOTE: I haven't been able to reproduce problem 2. unfortunately.

### DIFF
--- a/storm/src/main/storm/mesos/MesosSupervisor.java
+++ b/storm/src/main/storm/mesos/MesosSupervisor.java
@@ -40,8 +40,10 @@ import storm.mesos.shims.LocalStateShim;
 import storm.mesos.util.MesosCommon;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


drewrobb:
@erikdw this fixes problem 2 for me, thank you!

PR #123, opened 2016-04-09
document the purpose of the default storm & mesos version properties …

erikdw:
…in pom.xml


erikdw:
Just gonna merge this minor comment change.  Oh, the power that I wield! ;-)

PR #124, opened 2016-04-09
fix issue with vagrant start-nimbus.sh script

erikdw:
After we increased the package version name with all the useful info about
mesos and storm version, the process names are now too long for us to be
able to see the whole command in `ps` output (Linux limits the process length
to 4096 bytes by default).  So we cannot search for the correct daemon name
since it is cut off.  The brutal workaround is to kill all processes with
'storm' in their name.


erikdw:
Going to just merge this since it's a trivial change just for the vagrant setup.

erikdw:
The build failed because I deleted the branch prior to it completing.  The build doesn't mess with the vagrant setup at all, so this change cannot break the build.

PR #136, opened 2016-05-11
Fix Nimbus Crash on Bad Config

JessicaLHartog:
This pull request fixes an issue where if a user submitted a topology with `topology.mesos.cpu` or `topology.mesos.mem.mb` set to an incorrect type, the Nimbus will crash. Because of this, even when run under supervision the topology with the bad config is still living in ZK and continually kills the Nimbus.

To fix this, we now make the following changes:
1. Check to see if `topology.mesos.cpu` and `topology.mesos.mem.mb` are of type Number. If they are not, reset them to the default.
2. While we're at it, we should also check to see if `topology.mesos.cpu` and `topology.mesos.mem.mb` are too low for Mesos, if they are, reset them to the default.
3. Add tests so that these changes (and most of the other functionality of MesosCommon) are being verified on build so that this cannot happen again.

### DIFF
--- a/storm/src/main/storm/mesos/util/MesosCommon.java
+++ b/storm/src/main/storm/mesos/util/MesosCommon.java
@@ -40,6 +40,8 @@ public class MesosCommon {
   public static final String WORKER_NAME_PREFIX_DELIMITER = "topology.mesos.worker.prefix.delimiter";
   public static final String MESOS_COMPONENT_NAME_DELIMITER = "topology.mesos.component.name.delimiter";
 
+  public static final double MESOS_MIN_CPU = 0.01;
+  public static final double MESOS_MIN_MEM_MB = 32;
   public static final double DEFAULT_WORKER_CPU = 1;
   public static final double DEFAULT_WORKER_MEM_MB = 1000;
   public static final double DEFAULT_EXECUTOR_CPU = 0.1;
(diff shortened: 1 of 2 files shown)


storm/src/main/storm/mesos/util/MesosCommon.java
```diff
@@ -134,14 +136,34 @@ public static Map getFullTopologyConfig(Map conf, TopologyDetails info) {
...
+    Object cpuObject = topologyConfig.get(WORKER_CPU_CONF);
+    double topologyWorkerCpu = DEFAULT_WORKER_CPU;
+    if (cpuObject != null && !(cpuObject instanceof Number)) {
+      LOG.warn("Topology {} has invalid mesos cpu configuration: {} has type {}, when it should be a Number", info.getId(), cpuObject, cpuObject.getClass());
+    } else if (cpuObject != null && (cpuObject instanceof Number)) {
+      topologyWorkerCpu = ((Number) cpuObject).doubleValue();
+      if (topologyWorkerCpu < MESOS_MIN_CPU) {
+        LOG.warn("Topology {} has invalid mesos cpu configuration: {} is below {} which is the mesos minimum", info.getId(), topologyWorkerCpu, MESOS_MIN_CPU);
+        topologyWorkerCpu = DEFAULT_WORKER_CPU;
+      }
```
DarinJ: nit: I'd add "setting to DEFAULT_WORKER_CPU", so the user will have a little more information.

JessicaLHartog: @DarinJ Do you think it's also worth noting that the default is set when `topology.mesos.cpu` and `topology.mesos.mem.mb` are not set at all?

DarinJ: I don't think it would hurt.

dsKarthick: +1

storm/src/main/storm/mesos/util/MesosCommon.java
```diff
@@ -134,14 +136,34 @@ public static Map getFullTopologyConfig(Map conf, TopologyDetails info) {
 
   public static double topologyWorkerCpu(Map conf, TopologyDetails info) {
```
dsKarthick: How about the following?

```
public static Number getScalarTopologyConf(Map conf, String configName) throws IllegalArgumentException {
    Object val = (conf.get(configName) == null) ? null : conf.get(configName);

    if (!(val instanceof Number) || (val == null)) {
      LOG.info("Invalid {} - Expected type {}, Actual type {}", configName, Number.class, val.getClass());
      return null;
    }
    return (Number) val;
  }

  public static double topologyWorkerCpu(Map conf, TopologyDetails info) {
    Map topologyConfig = getFullTopologyConfig(conf, info);
    Object tmp = getScalarTopologyConf(topologyConfig, WORKER_CPU_CONF);

    Double topologyWorkerCpu = (tmp == null) ? DEFAULT_WORKER_CPU : ((Number) tmp).doubleValue();

    if (topologyWorkerCpu < MESOS_MIN_CPU) {
        LOG.warn("Invalid mesos cpu configuration for topoplogy `{}': {} is below the minimum required {}", info.getId(), topologyWorkerCpu, MESOS_MIN_CPU);
        return DEFAULT_WORKER_CPU;
    }
    return topologyWorkerCpu;
  }

  public static double topologyWorkerMem(Map conf, TopologyDetails info) {
    Map topologyConfig = getFullTopologyConfig(conf, info);
    Object tmp = getScalarTopologyConf(topologyConfig, WORKER_MEM_CONF);

    Double topologyWorkerMem = (tmp == null) ? DEFAULT_WORKER_MEM_MB : ((Number) tmp).doubleValue();

    if (topologyWorkerMem < MESOS_MIN_MEM_MB) {
      LOG.warn("Invalid mesos mem configuration for topoplogy `{}' : {} is below the minimum required {}", info.getId(), topologyWorkerMem, MESOS_MIN_MEM_MB);
      return DEFAULT_WORKER_MEM_MB;
    }
    return topologyWorkerMem;
  }
```

storm/src/test/storm/mesos/MesosCommonTest.java
```diff
@@ -0,0 +1,306 @@
...
+    double expectedResult = MesosCommon.DEFAULT_WORKER_CPU;
+    assertEquals(result, expectedResult, 0.001);
+
+    // Test what happens when config is too small
+    double cpuConfig = MesosCommon.MESOS_MIN_CPU;
+    cpuConfig -= 0.001;
+    conf.put(MesosCommon.WORKER_CPU_CONF, cpuConfig);
+    result = MesosCommon.topologyWorkerCpu(conf, info);
+    expectedResult = MesosCommon.DEFAULT_WORKER_CPU;
+    assertEquals(result, expectedResult, 0.0001);
```
dsKarthick: Define `private static final DELTA = 0.0001` as a member and use it here?

storm/src/test/storm/mesos/MesosCommonTest.java
```diff
@@ -0,0 +1,306 @@
...
+    Map expectedResult = new HashMap<>();
+    expectedResult.put("TEST_NIMBUS_CONFIG", 1);
+    expectedResult.put("TEST_TOPOLIGY_CONFIG", 3);
+    expectedResult.put("TEST_TOPOLOGY_OVERRIDE", 4);
+
+    assertEquals(result, expectedResult);
+  }
+
+  @Test
+  public void testTopologyWorkerCpu() throws Exception {
```
dsKarthick: One more test case that I can think off `conf.put(MesosCommon.WORKER_CPU_CONF, cpuConfig);`

storm/src/test/storm/mesos/MesosCommonTest.java
```diff
@@ -0,0 +1,306 @@
...
+    conf = new HashMap<>();
+    conf.put(MesosCommon.WORKER_CPU_CONF, 1.5);
+    info = new TopologyDetails("t4", conf, new StormTopology(), 1);
+    result = MesosCommon.topologyWorkerCpu(nimbusConf, info);
+    expectedResult = 1.5;
+    assertEquals(result, expectedResult, 0.001);
+  }
+
+  @Test
+  public void testTopologyWorkerMem() throws Exception {
```
dsKarthick: Same as the comments for `testTopologyWorkerCpu()`

dsKarthick:
Overall, LGTM.

JessicaLHartog:
@dsKarthick I think I addressed all of your comments please 👀 when you get a chance.

JessicaLHartog:
@DarinJ I took your advice, setting the log level for the case where the config options are not set at all to be debug (because otherwise it is very noisy). Please give it a once-over when you get some time.

dsKarthick:
:shipit:

erikdw:
🚢 🇮🇹

DarinJ:
🚀

dsKarthick:
Merged

Issue #127, opened 2016-04-09
use mvn release plugin


erikdw:
To perform releases we still have some manual steps we go through.  For an example, see [this description](https://github.com/mesos/storm/issues/76#issuecomment-207854029) of how `v0.1.0` was released.

We should instead use the `mvn release` command to create new releases.

Things to consider:

1. the tag should have a `v` prefix, so might need some tweaking to the `mvn release` configs to support that, since the version in our `pom.xml` doesn't have the `v` prefix.  i.e., the tag should be `v0.1.0` even if the version in the `pom.xml` is `0.1.0`.
2. since we embed the version with a bunch of other version-related stuff (storm & mesos versions) in the built artifacts' names, we need to be careful as we tweak these version fields.
  * e.g., `storm-mesos-0.1.0-storm0.10.0-mesos0.27.2.tgz` for `v0.1.0`
  * maybe it's ok to have it be `storm-mesos-v0.1.0-storm0.10.0-mesos0.27.2.tgz` instead.  But then need to regenerate the older release artifacts to be consistent.
3. we may want to update the release description on the [project's releases page](https://github.com/mesos/storm/releases).  Notably, we don't need to create the releases from this page when using the `mvn release` plugin, since the release creation is kicked off by pushing a tag to `mesos/storm`'s master branch, and the `mvn release` plugin does that for us.

erikdw:
We fixed 1. & 2. already.  3. is a nice-to-have.  (to automatically update the description with a link to the changes between the tags)   I've filed that issue as #162.  So this issue (#127) is complete.

PR #139, opened 2016-05-25
Use maven's release plugin to ease creation of new storm-mesos releases

erikdw:
Fixes issue #127.

Usage:

```
mvn release:clean release:prepare
```

Testing done:

1. changed the SCM settings at top of pom.xml to point at my own repo (erikdw/storm-mesos)
2. did the release:
  * `mvn release:clean release:prepare`
3. reverted changes by finding previous commit and then reverted my branch and pushed to my remote master to reset it:
  * `git reset --hard 0ed38ad ; git push -f origin master`
    * That is very dependent on what the previous commit SHA was... use with caution.
    * Also is done with the assumption that my git remote with the name of "origin" is pointing at my own erikdw/storm-mesos repo. 
4. reverted the tags locally and in my remote repo:
  * `git tag -d v0.1.3 ; git push --delete origin v0.1.3`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
-  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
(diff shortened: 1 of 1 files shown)


erikdw:
@DarinJ : can you please take a look when you have a chance?

DarinJ:
@erikdw sure - probably some time late today.  Looks good at first glance, just want to verify the interaction with build-release.sh.  I think we may want to consider eventually pulling some (if not all) of it's tasks into mvn, thoughts?

erikdw:
@DarinJ : thanks!  It *feels* right to do more stuff in maven directly, but it's also kind of painful so *might* not be worth the effort to pull *everything* from `build-release.sh` in to the pom files.  Let's mull it over, it seems like we could use the assembly plugin to do at least some of the `build-release.sh` logic.

DarinJ:
@erikdw checked it out.  LGTM.  Tested with my repo just to be sure.

👍 🚀:shipit: 🚢

erikdw:
thanks!

PR #142, opened 2016-06-03
Set the MesosSupervisor's storm.local.dir config to prevent conflicts between topologies

erikdw:
Without this change, we observed one supervisor killing all other topology's workers on a host,
which was being triggered because we were explicitly setting the MesosNimbus's storm.local.dir
to a hardcoded path.  If you don't set storm.local.dir in MesosNimbus then you would not be
affected by this problem.
    
See the comments in the code for the details on how this happens.


erikdw:
@dsKarthick : please review at your convenience (and to unblock you!).

DarinJ:
Lgtm

dsKarthick:
:shipit:

erikdw:
Thanks for the reviews @DarinJ & @dsKarthick.  Merging.

Issue #140, opened 2016-05-31
Storm 0.10 Thrift port type bug - Outdated docker image


jaimelive:
A bug interpreting the thrift port as a long instead of an int has been fixed by storm as seen in https://issues.apache.org/jira/browse/STORM-1645
however the docker images at mesos/storm and mesosphere/storm are still outdated.

erikdw:
@jaimelive : so you're pointing out that we should upgrade the storm version from 0.10.0 to 0.10.1 within the project.  Totally agreed!  I'll handle that.

erikdw:
@jaimelive : phew, finally the `mesos-0.28.2` packages were available and the builds passed.  So the fixed packages and docker images for v0.1.3 are here:
* https://hub.docker.com/r/mesos/storm/tags/
* https://github.com/mesos/storm/releases

PR #143, opened 2016-06-04
update storm to 0.10.1 and mesos to 0.28.2

erikdw:
Updating from storm-0.10.0 to storm-0.10.1 fixes issue #140, which references
the inability to change the Storm Thrift port in Storm version 0.10.0:
* https://issues.apache.org/jira/browse/STORM-1645

Also update from mesos-0.28.0 to mesos-0.28.1 since it has been released.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,9 +24,9 @@ script:
 matrix:
   fast_finish: true
   include:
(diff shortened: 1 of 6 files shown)


erikdw:
mesos-0.28.2 was just released an hour ago.  So updating our files to that.

PR #149, opened 2016-06-17
update the DOCKER_REPO variable to be the full repo path

erikdw:
This change avoids the assumption that the target docker repo will be `foo/storm`,
where only the `foo` portion is controlled by `DOCKER_REPO`.
With this fix it could be `foo/storm-mesos` for example.

The current `DOCKER_REPO` variable isn't referring to the full repo "slug", instead
it just references the user/organization portion of the slug, and assumes the project
name will always be `storm`.  That is not necessarily the case, for example in my fork
of `mesos/storm` I call the project `erikdw/storm-mesos` instead of just `erikdw/storm`.
That is done to avoid a conflict with the actual storm project (`apache/storm`), which
I also have forked under my name (`erikdw/storm`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ services:
 env:
   global:
     - TRAVIS_DOCKER_VERSION=1.10.3-0~trusty
-    - DOCKER_REPO=mesos
+    - DOCKER_REPO=mesos/storm
 before_install:
   - apt-cache madison docker-engine
   - travis_retry sudo apt-get -o Dpkg::Options::="--force-confnew" install -y docker-engine=${TRAVIS_DOCKER_VERSION}
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@
 #
 #
 RELEASE ?= `grep -1 -A 0 -B 0 '<version>' pom.xml | head -n 1 | awk '{print $1}' | sed -e 's/.*<version>//' | sed -e 's/<\/version>.*//'`
-DOCKER_REPO ?= mesos
+DOCKER_REPO ?= mesos/storm
 MIRROR ?=
 STORM_URL ?=
 JAVA_PRODUCT_VERSION ?= 7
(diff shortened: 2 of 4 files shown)


erikdw:
@salimane : can you please 👀 this?


FYI, I'm gonna send a follow-on change that I'd like your input on:

##### Remove hardcoded GitHub & DockerHub repo names for deploy, in favor of the $TRAVIS_REPO_SLUG environment variable:
```
diff --git .travis.yml .travis.yml
index f25e878..c3589e6 100644
--- .travis.yml
+++ .travis.yml
@@ -9,7 +9,7 @@ services:
 env:
   global:
     - TRAVIS_DOCKER_VERSION=1.10.3-0~trusty
-    - DOCKER_REPO=mesos/storm
+    - DOCKER_REPO=$TRAVIS_REPO_SLUG
 before_install:
   - apt-cache madison docker-engine
   - travis_retry sudo apt-get -o Dpkg::Options::="--force-confnew" install -y docker-engine=${TRAVIS_DOCKER_VERSION}
@@ -44,4 +44,4 @@ deploy:
   skip_cleanup: true
   on:
     tags: true
-    repo: mesos/storm
+    repo: $TRAVIS_REPO_SLUG
```

The advantage of is that we can do testing of build-related changes in child forks of `mesos/storm` without having to carefully update `deploy.on.repo` & `DOCKER_HUB` to our own personal repos.  Similarly we won't have to maintain such changes in our child forks (e.g., by doing `git stash save` and then `git stash pop` when we are working in that child repo).

(I've already tested a variant of this proposal, where I removed the `DOCKER_REPO` variable from `.travis.yml`, and set `DOCKER_REPO` to `$TRAVIS_REPO_SLUG` in the travis settings for my own repo.  And that worked.  So I'm pretty sure the above proposal will work too.  Just gotta test it.

JessicaLHartog:
LGTM 👍

erikdw:
Merging away!

PR #146, opened 2016-06-06
Schedule And Launch workers on consolidated offers

dsKarthick:
@brndnmtthws @DarinJ @erikdw @JessicaLHartog Please 👀 

Better version of https://github.com/mesos/storm/pull/111. 

Adresses following problems:
* Offers get too fragmented. So even though there are enough resources, because they are fragmented across multiple offers, they are not useful
* Some times, offer that is available at the time of allSlotsAvailableForScheduling gets removed from the rotating map before the storm-core invokes schedule to actually schedule the offers. This leads to a race-condition where some of the n workers take a long time to launch.

PR #151, opened 2016-06-21
use erikdw/storm's patched version of storm 0.9.6 to build storm-mesos package

erikdw:
Per the [release notes in erikdw/storm](https://github.com/erikdw/storm/releases) there are
a few tweaks that have been made to storm to make it more friendly for running storm-mesos:

1. Per-topology worker logs
2. Shorten java classpaths in various commands to ensure ps output includes full command
3. Flush bin/storm output output before exec'ing java process

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,8 +26,8 @@ matrix:
   include:
   - env: STORM_RELEASE="0.10.1" MESOS_RELEASE="0.28.2"
   - env: STORM_RELEASE="0.10.1" MESOS_RELEASE="0.27.2"
-  - env: STORM_RELEASE="0.9.6" MESOS_RELEASE="0.28.2"
-  - env: STORM_RELEASE="0.9.6" MESOS_RELEASE="0.27.2"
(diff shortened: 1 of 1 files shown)


erikdw:
@JessicaLHartog : please 👀 when you have a chance.  This has been tested in erikdw/storm-mesos, which is why I had to fix #149 & #150 first, in order to actually test this. :)

erikdw:
Once this is reviewed, I will release the next version of this project.  Then when #146 lands we will cut *another* version.  That should then free us to do the branch split that will allow support for storm 1.0 to precede.

JessicaLHartog:
LGTM, but out of sheer curiosity, why is the file name `apache-storm-0.9.6-storm-mesos2.tar.gz` and not simply `apache-storm-0.9.6-storm-mesos.tar.gz`?

erikdw:
'cause it's the 2nd release. :)   We'll do more if we need to iterate more.

This comes from:
* https://github.com/erikdw/storm/releases

JessicaLHartog:
Makes sense, thanks for clarifying. :shipit:

PR #154, opened 2016-06-23
[NEW] Schedule workers on consolidated offers

JessicaLHartog:
Since @dsKarthick is unavailable and the `master` branch is fundamentally broken, I am taking over #146 in his absence.

At the moment, this PR includes: 
* All of the content in #146 
* A change from @erikdw that [he requested](https://github.com/dsKarthick/storm-mesos/pull/1) 
* An additional commit related to [a comment](https://github.com/mesos/storm/pull/146#discussion-diff-67930452) I put on #146 that has not yet been addressed
* A commit adding in some comments as requested in #146 
* A commit that adds an assert for AggregatedOffers to make it a little more safe
* A re-write of the RangeResource `reserveAndGet` method to prevent us from modifying a list we are iterating over


storm/src/main/storm/mesos/resources/AggregatedOffers.java
```diff
@@ -64,6 +64,8 @@ public String getHostName() {
   }
 
   public void add(Protos.Offer offer) {
+    // We are unable to aggregate offers if they are from different workers
```
dsKarthick: different ~~workers~~ slaves

erikdw:
🚢 🇮🇹 

🏁 😓

salimane:
@erikdw does these changes applies to storm 0.9.6 shim ?

erikdw:
@salimane : yup, this is a fundamental change to the core scheduling logic which applies to both 0.9.6 & 0.10.x.   The shims are just slim "adapters" that allow the same storm-mesos codebase to work with different versions of storm's LocalState class.

PR #156, opened 2016-07-08
improve handling of status updates

erikdw:
1. For unrecognized status types we were falling into the default case and doing nothing.
  That was really bad with the introduction of TASK_ERROR for situations where the
  ExecutorInfo was mismatched between tasks for a topology, since the update/error wasn't
  appearing in the nimbus logs, only in the mesos-master logs.
  So now we log such updates at warning level.
2. Also just be more explicit about printing out the status updates for the normal flow,
  by logging at info level for TASK_RUNNING and the terminal states.
3. Also rename the `taskLost` method to `taskTerminated`, since it is not just for the
  lost statuses that we perform the action.

### DIFF
--- a/storm/src/main/storm/mesos/MesosNimbus.java
+++ b/storm/src/main/storm/mesos/MesosNimbus.java
@@ -300,7 +300,7 @@ public class MesosNimbus implements INimbus {
     }
   }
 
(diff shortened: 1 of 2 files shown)


JessicaLHartog:
👍 LGTM

PR #157, opened 2016-07-08
bin/storm-mesos launcher script: use exec to replace process instead of spawning separate child process

erikdw:
Mimic the behavior of storm's `bin/storm` launcher by using `exec`-style of replacing
the current process with the child.

This is useful when using a system like runit to supervise the daemons, since otherwise
there are 2 separate processes (the python launcher & the java process) and runit only
kills the 1st one.


JessicaLHartog:
👍 LGTM, love these simple short PRs

PR #167, opened 2016-08-14
Change LocalFileServer to allow respecting storm.yaml's nimbus.host

erikdw:
Prior to this the only way to configure the URL that LocalFileServer listens
to is via MESOS_NIMBUS_HOST environment variable.  That differs from the
behavior of the core MesosNimbus which supports that env variable, as well
as the nimbus.host field in the storm.yaml config.  So update the code to
make them equally configurable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,10 @@ services:
 - docker
 env:
   global:
-  - TRAVIS_DOCKER_VERSION=1.10.3-0~trusty
   - DOCKER_REPO=$TRAVIS_REPO_SLUG
 before_install:
 - apt-cache madison docker-engine
-- travis_retry sudo apt-get -o Dpkg::Options::="--force-confnew" install -y docker-engine=${TRAVIS_DOCKER_VERSION}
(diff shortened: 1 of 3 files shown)


erikdw:
@drewrobb : can you please review when you have a chance?  Thanks!

drewrobb:
LGTM!

drewrobb:
looks like build is failing for unrelated reasons

erikdw:
@drewrobb : thanks for the review.  Yeah, looks like the hardcoded `docker-engine` version is at fault, I removed it in another commit that I suppose I'll just piggy-back into this PR. My first build worked just fine:   https://travis-ci.org/mesos/storm/builds/152560889
Another build is running right now because I updated the commit description.

zhanghangchina:
@erikdw storm on mesos(0.28.2) cannot support zookeeper,so the mesos-master isn't HA!

erikdw:
@zhanghangchina : this isn't the right place to make that comment @zhanghangchina.  Please open an issue if you have a problem.

PR #179, opened 2016-11-10
Removed unused LRU taskIDtoOfferMap and related code

srishtyagrawal:
@erikdw @JessicaLHartog 

Using the following steps to run example storm topologies on vagrant setup for testing the changes:

1. `sudo sh -c 'echo "192.168.50.101 master" >> /etc/hosts'`
` sudo sh -c 'echo "192.168.50.102 slave"  >> /etc/hosts' `

2. Update storm.yaml
```
mesos.master.url: "zk://master:2181/mesos"
storm.zookeeper.servers:
- "master"
mesos.executor.uri: "file:///usr/local/storm/storm-mesos-0.1.8-SNAPSHOT-storm0.9.6-mesos0.27.0.tgz"
```
3. Run `bin/build-release.sh`
4. Run `mvn clean package` from local machine's storm starter folder (`_release/storm-mesos*/examples/storm-starter`)
5. Run `vagrant up`, then `vagrant provision` if asked for
6. Check UI at `master:8080` and manually run `sudo /vagrant/vagrant/start-nimbus.sh` from vagrant master.
7. Run the following commands from the target folder at vagrant master (`/vagrant/_release/storm-mesos*/examples/storm-starter/target`)
```
/vagrant/_release/storm-mesos-0.1.8-SNAPSHOT-storm0.9.6-mesos0.27.0/bin/storm jar storm-starter-0.9.6-jar-with-dependencies.jar storm.starter.ExclamationTopology ExclamationTopology
/vagrant/_release/storm-mesos-0.1.8-SNAPSHOT-storm0.9.6-mesos0.27.0/bin/storm jar storm-starter-0.9.6-jar-with-dependencies.jar storm.starter.WordCountTopology WordCountTopology
```
After executing the steps above I am able to see the topologies running and emitting tuples.

### DIFF
--- a/storm/src/main/storm/mesos/MesosNimbus.java
+++ b/storm/src/main/storm/mesos/MesosNimbus.java
@@ -108,7 +108,6 @@ public class MesosNimbus implements INimbus {
   public static final String CONF_MESOS_SECRET_FILE = "mesos.framework.secret.file";
 
   public static final String CONF_MESOS_CHECKPOINT = "mesos.framework.checkpoint";
-  public static final String CONF_MESOS_OFFER_LRU_CACHE_SIZE = "mesos.offer.lru.cache.size";
   public static final String CONF_MESOS_OFFER_FILTER_SECONDS = "mesos.offer.filter.seconds";
   public static final String CONF_MESOS_OFFER_EXPIRY_MULTIPLIER = "mesos.offer.expiry.multiplier";
   public static final String CONF_MESOS_LOCAL_FILE_SERVER_PORT = "mesos.local.file.server.port";
@@ -130,9 +129,6 @@ public class MesosNimbus implements INimbus {
   private Optional<Integer> _localFileServerPort;
   private RotatingMap<OfferID, Offer> _offers;
   private LocalFileServer _httpServer;
-  private Map<TaskID, Offer> taskIDtoOfferMap;
-  private ScheduledExecutorService timerScheduler =
-      Executors.newScheduledThreadPool(1);
   private IMesosStormScheduler _mesosStormScheduler = null;
 
   private boolean _preferReservedResources = true;
@@ -268,15 +264,6 @@ public class MesosNimbus implements INimbus {
         }
     );
 
-    Number lruCacheSize = Optional.fromNullable((Number) mesosStormConf.get(CONF_MESOS_OFFER_LRU_CACHE_SIZE)).or(1000);
-    final int intLruCacheSize = lruCacheSize.intValue();
-    taskIDtoOfferMap = Collections.synchronizedMap(new LinkedHashMap<Protos.TaskID, Protos.Offer>(intLruCacheSize + 1, .75F, true) {
-      // This method is called just after a new entry has been added
-      public boolean removeEldestEntry(Map.Entry eldest) {
-        return size() > intLruCacheSize;
-      }
-    });
-
     Number offerExpired = Optional.fromNullable((Number) mesosStormConf.get(Config.NIMBUS_MONITOR_FREQ_SECS)).or(10);
     Number expiryMultiplier = Optional.fromNullable((Number) mesosStormConf.get(CONF_MESOS_OFFER_EXPIRY_MULTIPLIER)).or(2.5);
     _timer.scheduleAtFixedRate(new TimerTask() {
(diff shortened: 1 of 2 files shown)


storm/src/main/storm/mesos/MesosNimbus.java
```diff
@@ -268,15 +265,6 @@ public void expire(Protos.OfferID key, Protos.Offer val) {
         }
     );
 
-    Number lruCacheSize = Optional.fromNullable((Number) mesosStormConf.get(CONF_MESOS_OFFER_LRU_CACHE_SIZE)).or(1000);
```
erikdw: `CONF_MESOS_OFFER_LRU_CACHE_SIZE` is no longer needed, so please remove that too.

storm/src/main/storm/mesos/MesosNimbus.java
```diff
@@ -130,9 +130,6 @@
   private Optional<Integer> _localFileServerPort;
   private RotatingMap<OfferID, Offer> _offers;
   private LocalFileServer _httpServer;
-  private Map<TaskID, Offer> taskIDtoOfferMap;
-  private ScheduledExecutorService timerScheduler =
```
erikdw: Good catch with `timerScheduler` no longer being used.

storm/src/main/storm/mesos/MesosNimbus.java
srishtyagrawal: Missed that one. Removed it now.

storm/src/main/storm/mesos/NimbusScheduler.java
```diff
@@ -102,7 +102,6 @@ public void statusUpdate(SchedulerDriver driver, TaskStatus status) {
       case TASK_LOST:
       case TASK_ERROR:
         LOG.info("Received status update: {}", taskStatusToString(status));
```
srishtyagrawal: TASK_ERROR has now the same action as TASK_RUNNING, so is TASK_STARTING. Isn't it misleading to have state of the task set to TASK_RUNNING when it has an error (TASK_ERROR) or is starting (TASK_STARTING)? This state is being set by setState() and I do not see state ever being set to TASK_ERROR or TASK_STARTING. Wondering if TASK_ERROR and TASK_STARTING are unused and untouched states.

```
12:58 PM sagrawal🐾  mesos-storm [master]$: git grep setState
storm/src/main/storm/mesos/MesosSupervisor.java:        .setState(TaskState.TASK_FINISHED)
storm/src/main/storm/mesos/MesosSupervisor.java:            .setState(TaskState.TASK_FAILED)
storm/src/main/storm/mesos/MesosSupervisor.java:          .setState(TaskState.TASK_RUNNING)
```

JessicaLHartog: Perhaps it would make sense to combine `TASK_RUNNING`, `TASK_FINISHED`, and `TASK_KILLED` into an info log line (as those are usually expected), but `TASK_LOST`, `TASK_FAILED`, and `TASK_ERROR` can use a warn log line instead?

Thoughts @erikdw @srishtyagrawal ?

erikdw: @srishtyagrawal 

> should I remove TASK_ERROR and TASK_STARTING?

Nope, perhaps I need to give more info:

1. These status updates can also come from the mesos-agent/slave & mesos-master (maybe?), not *just* from `MesosSupervisor`.  So you not seeing them being sent from the `MesosSupervisor` is not proof that they won't happen.
2. If you just removed those cases, then when we did receive those status updates we would fall through into the `Received unrecognized status` default case.

erikdw: @JessicaLHartog 

> `TASK_LOST`, `TASK_FAILED`, and `TASK_ERROR` can use a warn log line

`TASK_LOST` *used* to be a normal terminal state for our tasks.  It doesn't *seem* to be anymore, and I'm not sure about the reason.  So the last 2 definitely deserve WARN level logs, but I'm not sure about `TASK_LOST`, need to think more.

srishtyagrawal: @erikdw Thanks for the explanation, I see a lot of references to `TASK_ERROR` and `TASK_RUNNING` in the Apache Mesos code.

@JessicaLHartog, @erikdw do you mean changing the existing LOG.info for TASK_ERROR to LOG.warn and introducing logging for TASK_FAILED, TASK_KILLED etc (if they don't have logging already).

erikdw: Looks good mostly -- but with 3 identical lines I'm think we can construct the message before the switch and then just log it.  i.e., `String msg = String.format(...);`

erikdw:
@srishtyagrawal : looks good to me, thanks!  I'll merge this, and then I'll backport it to `storm-0.x` branch.

PR #180, opened 2016-11-25
Added suppressOffers for framework if its nimbus is not leader

simsasg:
@erikdw 

We want to run 3 HA nimbus instances. Mesos will see them as 3 frameworks.

Mesos master generates offers to all registered frameworks using Dominant Resource Fairness algorithm.
Offers are held by storm-mesos framework even if nimbus is not leader and this can cause resource starvation in cluster, because remaining offers will be held by other (non leader) storm frameworks.

The solution is to suppressOffers for non leader frameworks and reviveOffers only for framework with nimbus leader. If new nimbus leader will be elected,  this task would reviveOffers for new leader.

There are 2 new configuration settings:

* `mesos.nimbus.leader.check.delay.secs`: Nimbus leader check task delay in seconds. Defaults to "120". If this setting is set to 0 - all frameworks will be suppressed at first and will reviveOffers after  `mesos.nimbus.leader.check.period.secs`. You might want to set this setting to bigger value if all active  frameworks should accept offers at first and only suppress offers after this delay.
* `mesos.nimbus.leader.check.period.secs`: Nimbus leader check task period in seconds. Defaults to "60".

I also removed adding Config.NIMBUS_HOST to config because it was deprecated in storm 1.0.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,10 +150,11 @@ For local development and familiarizing yourself with Storm/Mesos, please see th
 * `mesos.disallowed.hosts`: Disallowed hosts to run topology, which takes hostname list as a back list.
 * `mesos.framework.role`: Framework role to use. Defaults to "*".
 * `mesos.framework.checkpoint`: Enabled framework checkpoint or not. Defaults to false.
-* `mesos.offer.lru.cache.size`: LRU cache size. Defaults to "1000".
 * `mesos.offer.filter.seconds`: Number of seconds to filter unused Mesos offers. These offers may be revived by
   the framework when needed. Defaults to "120".
 * `mesos.offer.expiry.multiplier`: Offer expiry multiplier for `nimbus.monitor.freq.secs`. Defaults to "2.5".
+* `mesos.nimbus.leader.check.delay.secs`: Nimbus leader check task delay in seconds. Defaults to "120".
+* `mesos.nimbus.leader.check.period.secs`: Nimbus leader check task period in seconds. Defaults to "60".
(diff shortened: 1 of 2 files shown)


DarinJ:
I think this is the wrong approach.  As apposed to registering three minibuses as different frameworks we should use mesos's ha features.  So each nimbus would need to register w/ the same framework Id.  This will of course mean we need to store the framework Id in a globally accessible spot (likely zookeeper) vs local state as we do presently.

erikdw:
Agree with @DarinJ.

BTW, just to state it explicitly, there is another major consideration if you want tasks to survive Nimbus failover:  you must advertise the same `host:port` value for the Jetty server embedded in MesosNimbus that serves the `storm.yaml` config to the executors.  i.e., that value is embedded as a URI in the `ExecutorInfo`, which must be the same for all tasks.  If the ExecutorInfo varies then you'll get task launching failures which don't bubble up to any UI and thus are hard to notice.  Doing that in our setup is easy, since we use a dedicated server for the Nimbus and just have a DNS alias that we flip.  We also set `mesos.local.file.server.port` to a static value, so when we restart the MesosNimbus it's the same port.  However, I'm unsure of how to handle this scenario when using Mesos to launch even 1 Nimbus, much less 2+ of them.  I remember @DarinJ commenting that using MesosDNS was an approach, but I am wholly ignorant of that technology at this moment.

simsasg:
Thanks for feedback. I agree, registering one mesos framework is better approach. I will close this pull request.

PR #189, opened 2017-01-02
Change deprecated `nimbus.host`-> `nimbus.seeds`

eyalzek:
A change introduced in v1.0.0 of apache storm. Using `nimbus.host` will still work but will produces the following error:
`Using deprecated config nimbus.host for backward compatibility. Please update your storm.yaml so it only has config nimbus.seeds`
https://issues.apache.org/jira/browse/STORM-1185

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,7 +224,7 @@ to Nimbus, you'll need to know the Thrift host and API port. In the Marathon exa
 one assigned by Marathon. For example, if the host is `10.0.0.1` and second port is `32001`, run:
 
 ```
-$ ./bin/storm jar -c nimbus.host=10.0.0.1 -c nimbus.thrift.port=32001 examples/storm-starter/storm-starter-topologies-0.9.6.jar storm.starter.WordCountTopology word-count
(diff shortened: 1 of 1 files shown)


erikdw:
@eyalzek : have you tested this suggested change?  I don't think it's so easy, the MesosNimbus code makes assumptions that it can use `nimbus.host` to drive various behaviors, and because of the change from single string to an array, it's not *quite* as simple as changing the code to look at the alternate parameter name.

eyalzek:
Yes, that is indeed a warning and not an error, my bad.
I changed this setting on my `storm.yaml` to get rid of these warning and it's working ok. Since I've seen that on the README here it's still referencing `nimbus.host` I figured it should be updated. I might not be aware of the full consequences of it though.

erikdw:
@eyalzek : ahh, yep, it's a *bit* more nuanced.  Not rocket science or anything, but requires a bit of effort to change the code too.  I've filed issue #190 for fixing MesosNimbus to support `nimbus.seeds`, so I'm gonna close this PR.

PR #191, opened 2017-02-15
Exclude examples dir from release packages

fuji-151a:
I could reduced the capacity of release packages from 177MB to 114MB by excluding `examples/` directory from release packages.

See #178 for details.

Please check this pull request.

Thanks.

PR #193, opened 2017-02-16
improve README for running an example topology under Marathon

erikdw:


erikdw:
@JessicaLHartog : this should assuage your concerns with the removal of the `examples/` dir in #191.

JessicaLHartog:
LGTM :shipit:

erikdw:
thanks for the review @JessicaLHartog !

PR #194, opened 2017-02-16
exclude storm's external libraries from storm-mesos release tarball

erikdw:


erikdw:
@JessicaLHartog : please review.

JessicaLHartog:
LGTM :shipit:

PR #195, opened 2017-02-28
Implements to register framework user

fuji-151a:
I Implemented to be able to register framework user to run with Mesos.
Please check this pull request.

Thanks.


erikdw:
@fuji-151a : have you tested that this works?  I'm using mesos-1.0.1 and storm-1.0.2.  When I set the `mesos.framework.user` to an alternate username, the `MesosSupervisor` *is* launched under that user, however the subsequent `Worker` processes are still owned by root.

I'm assuming we'd need to set some config in storm itself.  Do you know what that is?  We should update the docs if that's the case.

erikdw:
Actually, I think I was wrong.  It didn't make sense to me that a non-root process (`MesosSupervisor`) would be allowed to launch a root process (`Worker`).  And upon further inspection, I think somehow I had some lingering processes from earlier runs and those misled me.  So I *believe* this is working correctly... I'll play around a bit more before merging.

erikdw:
Yup, this looks good to me, thanks for the contribution @fuji-151a !

PR #198, opened 2017-06-06
Renamed a few classes for clarity: DefaultScheduler and NimbusScheduler

JessicaLHartog:
* `DefaultScheduler` was renamed to `StormSchedulerImpl` since this class is a Scheduler Implementation that mimics Storm's default scheduling behavior
   * Also renamed `DefaultSchedulerTest` to reflect these changes
* `NimbusScheduler` was renamed to `NimbusMesosScheduler` since it does not actually schedule things on behalf of Nimbus but acts as an intermediary between Nimbus and Mesos

### DIFF
--- a/storm/src/main/storm/mesos/MesosNimbus.java
+++ b/storm/src/main/storm/mesos/MesosNimbus.java
@@ -125,7 +125,7 @@ public class MesosNimbus implements INimbus {
   private Optional<Integer> _localFileServerPort;
   private RotatingMap<OfferID, Offer> _offers;
   private LocalFileServer _httpServer;
-  private IMesosStormScheduler _mesosStormScheduler = null;
+  private IMesosStormScheduler _stormScheduler = null;
 
   private boolean _preferReservedResources = true;
   private Optional<String> _container = Optional.absent();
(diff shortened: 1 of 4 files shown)


storm/src/main/storm/mesos/MesosNimbus.java
```diff
@@ -116,7 +116,7 @@
   private final Object _offersLock = new Object();
   protected java.net.URI _configUrl;
   private LocalStateShim _state;
-  private NimbusScheduler _scheduler;
+  private NimbusMesosScheduler _scheduler;
```
erikdw: maybe rename the variable from `_scheduler` to `_mesosScheduler`?

storm/src/main/storm/mesos/MesosNimbus.java
```diff
@@ -140,7 +140,7 @@ private static Set listIntoSet(List l) {
   }
 
   public MesosNimbus() {
-    this._mesosStormScheduler = new DefaultScheduler();
+    this._mesosStormScheduler = new StormSchedulerImpl();
```
erikdw: maybe rename the variable from `_mesosStormScheduler` to `_stormScheduler`?

erikdw:
Nice, I find this a lot more intuitive.  :shipit:

PR #207, opened 2017-08-09
Fix travis-ci build

erikdw:
Many of our latest PRs have failed their builds, as you can see here:
* https://travis-ci.org/mesos/storm/pull_requests

This PR fixes 2 issues that were breaking the build.


erikdw:
thanks!

PR #209, opened 2017-08-15
Automated logviewer launch for 0.x

changreytang:
#### Notes
- Automated logviewer launch on any worker hosts running Storm bits (0.x specific)
- State of these logviewer daemons are tracked per host by stuffing state into ZooKeeper
- Reconciliation is also implemented to ensure that logviewers are brought back alive if ever killed.

#### ZKClient.java
- Wrapper around Apache Curator in order to interact with ZK
- Can create, delete, and check for existence of nodes in ZK
- Used to track which hosts are running logviewer daemons by tracking that state into ZK
- Junit tests are written for this class

#### MesosNimbus.java
- Added TimerTask to perform implicit reconciliation every 5 minutes
  - Implicit reconciliation reconciles that state of every running task between the framework scheduler and master (sends update of task status to *statusUpdate* method in *NimbusMesosScheduler*)
- *launchLogviewer* method called in *allSlotsAvailableForScheduling* loops through the *existingSupervisors* and schedules the logviewer daemon as Mesos tasks based on the offers available on each host
  - Checks ZK if logviewer already running on the host which the supervisor is running on, as such, ensures that logviewer only runs on hosts with Storm bits

#### NimbusMesosScheduler.java
- Updated *statusUpdate* method to look for logviewer tasks to check if they need to be relaunched
- *updateLogviewerState* method updates the state of the logviewer task in ZK if needed
  - *reviveOffers* if logviewer needs to be relaunched

#### storm.yaml
- Added new config called *storm.logviewer.zookeeper.dir* which allows you to configure under which "path" you want the logviewer state to be stored in ZK

### DIFF
--- a/storm/pom.xml
+++ b/storm/pom.xml
@@ -203,6 +203,16 @@
       <artifactId>storm-shim</artifactId>
       <version>${project.parent.version}</version>
     </dependency>
-
+    <dependency>
+      <groupId>org.apache.curator</groupId>
+      <artifactId>curator-recipes</artifactId>
+      <version>2.12.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.curator</groupId>
+      <artifactId>curator-test</artifactId>
+      <version>2.12.0</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 </project>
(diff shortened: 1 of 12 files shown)


erikdw:
@changreytang : thanks for updating this!

I'm confused by the presence of this commit in your PR: https://github.com/mesos/storm/pull/209/commits/7ae98cb358d73cd00489ae5f9b8c30aa76b1980f, can you explain why it's there or ideally get rid of it somehow?

The commit comment for it is from a different commit from awhile ago: https://github.com/mesos/storm/commit/4153391a0972918f1dbbb7a470d384968510aab2

changreytang:
@erikdw not 100% sure why it's like that, must've been something when I rebased, will look into it to see if I can get rid of it

erikdw:
This has been merged via #228, see that PR for details.

PR #212, opened 2017-08-16
Automated logviewer launch for 1.x

changreytang:
#### Notes
- Automated logviewer launch on any worker hosts running Storm bits (1.x specific)
- State of these logviewer daemons are tracked per host by stuffing state into ZooKeeper
- Reconciliation is also implemented to ensure that logviewers are brought back alive if ever killed.

#### ZKClient.java
- Wrapper around Apache Curator in order to interact with ZK
- Can create, delete, and check for existence of nodes in ZK
- Used to track which hosts are running logviewer daemons by tracking that state into ZK
- Junit tests are written for this class

#### MesosNimbus.java
- Added TimerTask to perform implicit reconciliation every 5 minutes
  - Implicit reconciliation reconciles that state of every running task between the framework scheduler and master (sends update of task status to *statusUpdate* method in *NimbusMesosScheduler*)
- *launchLogviewer* method called in *allSlotsAvailableForScheduling* loops through the *existingSupervisors* and schedules the logviewer daemon as Mesos tasks based on the offers available on each host
  - Checks ZK if logviewer already running on the host which the supervisor is running on, as such, ensures that logviewer only runs on hosts with Storm bits

#### StormSchedulerImpl.java
- Added an *offersRequestTracker* that tracks if any storm tasks (topologies) or sidecar tasks need offers. This ensures that offers are only suppressed when no tasks need them.

#### NimbusMesosScheduler.java
- Updated *statusUpdate* method to look for logviewer tasks to check if they need to be relaunched
- *updateLogviewerState* method updates the state of the logviewer task in ZK if needed
  - *checkRunningLogviewerState* method checks if a running Mesos tasks exists for logviewer that isn't tracked in ZK and updates ZK state if necessary

#### MesosCommon.java
- Added *getMesosFrameworkName* as an utility to get the framework name from the storm.yaml
- Modified *supervisorId* method to format the ID with the framework name

### DIFF
--- a/storm/src/main/storm/mesos/MesosNimbus.java
+++ b/storm/src/main/storm/mesos/MesosNimbus.java
@@ -17,6 +17,7 @@
  */
 package storm.mesos;
 
+import com.google.protobuf.InvalidProtocolBufferException;
 import org.apache.storm.Config;
 import org.apache.storm.scheduler.INimbus;
 import org.apache.storm.scheduler.IScheduler;
@@ -46,6 +47,7 @@ import org.apache.mesos.Protos.TaskStatus;
 import org.apache.mesos.Protos.Value.Range;
 import org.apache.mesos.Protos.Value.Ranges;
 import org.apache.mesos.Protos.Value.Scalar;
+import org.apache.mesos.Protos.TaskState;
 import org.apache.mesos.SchedulerDriver;
 import org.json.simple.JSONValue;
 import org.slf4j.Logger;
@@ -282,13 +284,28 @@ public class MesosNimbus implements INimbus {
     _timer.scheduleAtFixedRate(new TimerTask() {
       @Override
       public void run() {
-        // performing "implicit" reconciliation; master will respond with the latest state for all currently known
-        // non-terminal tasks in the framework scheduler's statusUpdate() method
+        // performing "explicit" reconciliation; master will respond with the latest state for all logviewer tasks
+        // in the framework scheduler's statusUpdate() method
         Collection<TaskStatus> taskStatuses = new ArrayList<TaskStatus>();
+        List<String> logviewerPaths = _zkClient.getChildren(_logviewerZkDir);
+        if (logviewerPaths == null) {
+          _driver.reconcileTasks(taskStatuses);
+          return;
+        }
+        for (String path : logviewerPaths) {
+          TaskID logviewerTaskId = TaskID.newBuilder()
+                                          .setValue(new String(_zkClient.getNodeData(String.format("%s/%s", _logviewerZkDir, path))))
+                                          .build();
+          TaskStatus logviewerTaskStatus = TaskStatus.newBuilder()
+                                                     .setTaskId(logviewerTaskId)
+                                                     .setState(TaskState.TASK_RUNNING)
+                                                     .build();
+          taskStatuses.add(logviewerTaskStatus);
+        }
         _driver.reconcileTasks(taskStatuses);
-        LOG.info("Performing task reconciliation between scheduler and master");
+        LOG.info("Performing task reconciliation between scheduler and master on following tasks: {}", taskStatuses.toString());
       }
-    }, TASK_RECONCILIATION_INTERVAL, TASK_RECONCILIATION_INTERVAL); // reconciliation performed every 5 minutes
+    }, 0, TASK_RECONCILIATION_INTERVAL); // reconciliation performed every 5 minutes
   }
 
   public void shutdown() throws Exception {
@@ -484,6 +501,7 @@ public class MesosNimbus implements INimbus {
 
       String logviewerZKPath = String.format("%s/%s", _logviewerZkDir, nodeId);
       _zkClient.createNode(logviewerZKPath);
+      _zkClient.updateNodeData(logviewerZKPath, taskId.getValue());
       LOG.info("launchLogviewer: Create logviewer state in zk: {}", logviewerZKPath);
     }
     if (needLogviewer) {
--- a/storm/src/main/storm/mesos/MesosNimbus.java
+++ b/storm/src/main/storm/mesos/MesosNimbus.java
@@ -42,6 +42,7 @@ import org.apache.mesos.Protos.OfferID;
 import org.apache.mesos.Protos.Resource;
 import org.apache.mesos.Protos.TaskID;
 import org.apache.mesos.Protos.TaskInfo;
+import org.apache.mesos.Protos.TaskStatus;
 import org.apache.mesos.Protos.Value.Range;
 import org.apache.mesos.Protos.Value.Ranges;
 import org.apache.mesos.Protos.Value.Scalar;
@@ -84,6 +85,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.Timer;
+import java.util.TimerTask;
 
 import static storm.mesos.util.PrettyProtobuf.offerIDListToString;
 import static storm.mesos.util.PrettyProtobuf.offerToString;
@@ -108,9 +111,13 @@ public class MesosNimbus implements INimbus {
   public static final String CONF_MESOS_CONTAINER_DOCKER_IMAGE = "mesos.container.docker.image";
   public static final String CONF_MESOS_SUPERVISOR_STORM_LOCAL_DIR = "mesos.supervisor.storm.local.dir";
   public static final String FRAMEWORK_ID = "FRAMEWORK_ID";
+  public static final String DEFAULT_MESOS_COMPONENT_NAME_DELIMITER = "|";
 
   public static final String CONF_ZOOKEEPER_SERVERS = "storm.zookeeper.servers";
   public static final String CONF_ZOOKEEPER_PORT = "storm.zookeeper.port";
+  public static final String CONF_STORM_LOGVIEWER_ZK_DIR = "storm.logviewer.zookeeper.dir";
+
+  public static final int TASK_RECONCILIATION_INTERVAL = 300000; // 5 minutes
 
   private static final Logger LOG = LoggerFactory.getLogger(MesosNimbus.class);
   private final Object _offersLock = new Object();
@@ -120,6 +127,7 @@ public class MesosNimbus implements INimbus {
   protected volatile SchedulerDriver _driver;
   private volatile boolean _registeredAndInitialized = false;
   private ZKClient _zkClient;
+  private String _logviewerZkDir;
   private Timer _timer = new Timer();
   private Map mesosStormConf;
   private Set<String> _allowedHosts;
@@ -219,13 +227,14 @@ public class MesosNimbus implements INimbus {
     if (zooKeeperPort == null || zooKeeperServers == null) {
       LOG.error("ZooKeeper configs are not found in storm.yaml");
     } else {
-      String connectionString = "";
+      StringBuilder connectionString = new StringBuilder();
       for (String server : zooKeeperServers) {
-        connectionString += server + ":" + zooKeeperPort + ",";
+        connectionString.append(String.format("%s:%s,", server, zooKeeperPort));
       }
       _zkClient = new ZKClient(connectionString.substring(0, connectionString.length() - 1));
-      if (!_zkClient.nodeExists("/logviewers")) {
-        _zkClient.createNode("/logviewers");
+      if (!_zkClient.nodeExists(_logviewerZkDir)) {
+        _zkClient.createNode(_logviewerZkDir);
+        LOG.info("Created general ZK directory for logviewer state at: {}", _logviewerZkDir);
       }
     }
 
@@ -235,7 +244,7 @@ public class MesosNimbus implements INimbus {
     }
 
     _container = Optional.fromNullable((String) conf.get(CONF_MESOS_CONTAINER_DOCKER_IMAGE));
-    _mesosScheduler = new NimbusMesosScheduler(this);
+    _mesosScheduler = new NimbusMesosScheduler(this, _zkClient, _logviewerZkDir);
 
     // Generate YAML to be served up to clients
     _generatedConfPath = Paths.get(
@@ -275,6 +284,17 @@ public class MesosNimbus implements INimbus {
 
     _state.put(FRAMEWORK_ID, id.getValue());
     _offers = new HashMap<Protos.OfferID, Protos.Offer>();
+
+    _timer.scheduleAtFixedRate(new TimerTask() {
+      @Override
+      public void run() {
+        // performing "implicit" reconciliation; master will respond with the latest state for all currently known
+        // non-terminal tasks
+        Collection<TaskStatus> taskStatuses = new ArrayList<TaskStatus>();
+        _driver.reconcileTasks(taskStatuses);
+        LOG.info("Performing tasking reconciliation between scheduler and master");
+      }
+    }, TASK_RECONCILIATION_INTERVAL, TASK_RECONCILIATION_INTERVAL); // reconciliation performed every 5 minutes
   }
 
   public void shutdown() throws Exception {
@@ -353,7 +373,9 @@ public class MesosNimbus implements INimbus {
       return new ArrayList<WorkerSlot>();
     }
     synchronized (_offersLock) {
-      launchLogviewer(existingSupervisors);
+      if (!_container.isPresent()) {
+        launchLogviewer(existingSupervisors);
+      }
       return _stormScheduler.allSlotsAvailableForScheduling(
               _offers,
               existingSupervisors,
@@ -388,9 +410,9 @@ public class MesosNimbus implements INimbus {
       List<TaskInfo> logviewerTask = new ArrayList<TaskInfo>();
       LOG.info("launchLogviewer: Supervisor ID: {}", supervisor.getId());
 
-      String nodeId = supervisor.getId().split("|")[0];
+      String nodeId = supervisor.getId().split("\\" + DEFAULT_MESOS_COMPONENT_NAME_DELIMITER)[0];
 
-      if (_zkClient.nodeExists(String.format("/logviewers/%s", nodeId))) {
+      if (_zkClient.nodeExists(String.format("%s/%s", _logviewerZkDir, nodeId))) {
         LOG.info("launchLogviewer: Logviewer already exists on this host: {}", nodeId);
         continue;
       }
@@ -422,7 +444,7 @@ public class MesosNimbus implements INimbus {
               .setValue(logviewerCommand);
 
       TaskID taskId = TaskID.newBuilder()
-              .setValue(String.format("%s-logviewer", nodeId))
+              .setValue(String.format("%s%slogviewer",  nodeId, DEFAULT_MESOS_COMPONENT_NAME_DELIMITER))
               .build();
 
       TaskInfo task = TaskInfo.newBuilder()
@@ -442,9 +464,9 @@ public class MesosNimbus implements INimbus {
         _offers.remove(offerID);
       }
 
-      String logviewerZKNodeName = String.format("/logviewers/%s", nodeId);
-      _zkClient.createNode(logviewerZKNodeName);
-      LOG.info("launchLogviewer: Updating logviewer state in zk: {}", logviewerZKNodeName);
+      String logviewerZKPath = String.format("%s/%s", _logviewerZkDir, nodeId);
+      _zkClient.createNode(logviewerZKPath);
+      LOG.info("launchLogviewer: Create logviewer state in zk: {}", logviewerZKPath);
     }
   }
 
@@ -812,9 +834,5 @@ public class MesosNimbus implements INimbus {
       credential = credentialBuilder.build();
     }
     return credential;
-  }
-
-  private void createLogviewer() {
-
   }
 }
--- a/storm/src/main/storm/mesos/MesosNimbus.java
+++ b/storm/src/main/storm/mesos/MesosNimbus.java
@@ -111,11 +111,7 @@ public class MesosNimbus implements INimbus {
   public static final String CONF_MESOS_CONTAINER_DOCKER_IMAGE = "mesos.container.docker.image";
   public static final String CONF_MESOS_SUPERVISOR_STORM_LOCAL_DIR = "mesos.supervisor.storm.local.dir";
   public static final String FRAMEWORK_ID = "FRAMEWORK_ID";
-  public static final String DEFAULT_MESOS_COMPONENT_NAME_DELIMITER = "|";
-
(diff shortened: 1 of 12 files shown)


storm/src/main/storm/mesos/MesosNimbus.java
```diff
@@ -107,13 +111,24 @@
   public static final String CONF_MESOS_CONTAINER_DOCKER_IMAGE = "mesos.container.docker.image";
   public static final String CONF_MESOS_SUPERVISOR_STORM_LOCAL_DIR = "mesos.supervisor.storm.local.dir";
   public static final String FRAMEWORK_ID = "FRAMEWORK_ID";
+  public static final String DEFAULT_MESOS_COMPONENT_NAME_DELIMITER = "|";
+
+  public static final String CONF_ZOOKEEPER_SERVERS = "storm.zookeeper.servers";
+  public static final String CONF_ZOOKEEPER_PORT = "storm.zookeeper.port";
+  public static final String CONF_STORM_LOGVIEWER_ZK_DIR = "storm.logviewer.zookeeper.dir";
```
erikdw: Lets shove `mesos` in there, since this config option is specific to our framework, and without that it looks to be a general storm configuration parameter.  e.g., `storm.mesos.logviewer.zookeeper.dir`.

erikdw: Here's another thought: right now there are basically 3 different dirs where state is stored related to StaaS:
1. `/<storm-something>/`:  this is where Storm itself stores some state.  It is configured by `storm.zookeeper.root` in `storm.yaml`.
2. `/<mesos-something>/`: this is where Mesos stores some state.
3. `/consumers/`: this is where we tell our users to store their Kafka consumer state (when they are using ZK).

So we're creating a 4th place.  My question is:  would it be beneficial to have a 4. `/<storm-mesos-something>` specific dir, and then have the logviewer dir be a subdir under that?   e.g., I can see us putting other kinds of state in here eventually.

storm/src/main/storm/mesos/NimbusMesosScheduler.java
```diff
@@ -109,6 +119,28 @@ public void statusUpdate(SchedulerDriver driver, TaskStatus status) {
...
+    String logviewerZKPath = String.format("%s/%s", logviewerZkDir, nodeId);
+    if (zkClient.nodeExists(logviewerZKPath)) {
+      LOG.info("updateLogviewerState: Remove logviewer state in zk: {}", logviewerZKPath);
+      zkClient.deleteNode(logviewerZKPath);
+      LOG.info("updateLogviewerState: Reviving offers for logviewer relaunch");
+      mesosNimbus._driver.reviveOffers();
+      StormSchedulerImpl stormScheduler = (StormSchedulerImpl) mesosNimbus.getForcedScheduler();
+      stormScheduler.unsetOffersSuppressed();
+    } else {
+      LOG.error("Task exists for logviewer that isn't tracked in ZooKeeper");
```
erikdw: Also, can you please print out the Task ID whenever you are emitting logs about a particular task?  I realize it might feel redundant with the other log that got printed right before this in the `statusUpdate()` method, however it's helpful to be able to individually grep lines in log files.  I really wish the implementors of storm-kafka had added more information into each log line, especially because many lines get interleaved and come from different spout threads running in the same worker process, so it's impossible to figure out which partition and/or topic is being referred to in various storm-kafka log lines.

storm/src/main/storm/mesos/util/MesosCommon.java
```diff
@@ -60,7 +60,7 @@
   public static final String SUPERVISOR_ID = "supervisorid";
   public static final String ASSIGNMENT_ID = "assignmentid";
   public static final String DEFAULT_WORKER_NAME_PREFIX_DELIMITER = "_";
-  public static final String DEFAULT_MESOS_COMPONENT_NAME_DELIMITER = " | ";
+  public static final String DEFAULT_MESOS_COMPONENT_NAME_DELIMITER = "|";
```
erikdw: Also, it seems like `DEFAULT_MESOS_COMPONENT_NAME_DELIMITER` & `getMesosComponentNameDelimiter()` weren't used anywhere except tests before your change, right?

It *used* to be used per my changes here:
* https://github.com/mesos/storm/commit/f7f18f86272c1f02e4574996a207e53ab599a8e8

However it seems those changes were destroyed later.  I *definitely* wanna leave the ` | ` styling for the component *names*.  However, for the IDs we might want or need to not have spaces around the delimiter.  I should also reapply my changes that made the name delimiter configurable, instead of hardcoding it to ` | ` as was done later (maybe by me...).

So I think it might be appropriate to use a different field (`MESOS_COMPONENT_ID_DELIMITER`), and have it hardcoded (not configurable) instead of reusing the *name* delimiter for this functionality.

storm/src/main/storm/mesos/util/MesosCommon.java
```diff
@@ -105,7 +105,7 @@ public static String taskId(String nodeid, int port) {
   }
 
   public static String supervisorId(String nodeid, String topologyId) {
-    return String.format("%s-%s", nodeid, topologyId);
+    return String.format("%s%s%s", nodeid, DEFAULT_MESOS_COMPONENT_NAME_DELIMITER, topologyId);
```
erikdw: Shouldn't this be a call to `MesosCommon.getMesosComponentNameDelimiter()` instead of using `DEFAULT_MESOS_COMPONENT_NAME_DELIMITER` directly?  I realize it might be annoying to propagate the `Map conf` and `TopologyDetails info` through this call, so it might not be worth it.

storm/src/main/storm/mesos/MesosNimbus.java
```diff
@@ -335,6 +382,92 @@ public boolean isHostAccepted(String hostname) {
...
+        continue;
+      }
+
+      CommandInfo.Builder commandInfoBuilder = CommandInfo.newBuilder()
+              .addUris(URI.newBuilder().setValue((String) mesosStormConf.get(CONF_EXECUTOR_URI)))
+              .addUris(URI.newBuilder().setValue(configUri))
+              .setValue(logviewerCommand);
+
+      TaskID taskId = TaskID.newBuilder()
+              .setValue(String.format("%s%slogviewer",  nodeId, DEFAULT_MESOS_COMPONENT_NAME_DELIMITER))
```
erikdw: e.g., of existing Task ID and Task Name for logviewer:
* Task ID: `worker-host4.dev|logviewer`
* Task Name: `logviewer`

And here's an existing Task ID and Task Name for a storm worker process:
* Task ID: `worker-host4.dev-31000-1503091419.181
* Task Name: `storm-worker | topology-foo-16-1493410791 | worker-host4.dev:31000`

Let's think about how to improve all of these for the situation where we're in a Blue-Green deployment situation with multiple Storm-on-Mesos frameworks running in the same Mesos cluster.  Goals:
1. Task ID *must* be unique or Mesos has major issues (see [this thread](https://mail-archives.apache.org/mod_mbox/mesos-user/201602.mbox/%3CCAO5KYW8%2BXMWc1dXtEo20BAsfGow028jwjL2ubMinP%2BK%2BvdOh8w@mail.gmail.com%3E) on the mesos user list where I posted an issue we encountered).
   * For worker processes that's handled by the combo of host + port, but also throwing in the timestamp of the launching of the task ensures it is unique.
   * For logviewer, it won't be unique if there is more than 1 Storm framework running.  We can make it unique by shoving in either the logviewer port and/or the Framework name.  More on this decision below.
2. Task Name should be meaningful.
   * For the worker process it's pretty useful (has the type of process, the topology ID and the slot ID), but it's lacking the framework name. e.g., `Storm v1.0.3 | storm-worker | ....`.   We can make that change with your PR or in a separate PR.
   * For the logviewer process it should at the minimum be enhanced to have:
        1. `storm-logviewer` instead of `logviewer`.
        2. ` | ` delimited strings with more information.
        3. The framework name.
        4. The logviewer port (from the storm config that is known to the MesosNimbus, it's `logviewer.port`).
    * So I'm thinking this would be good:  `<Framework Name> | storm-logviewer | port <logviewer.port>`.
       * e.g., `Storm v1.0.3 | storm-logviewer | port 8888`
3. Task ID appears in logs more than Task Name, so we should think about things we might wanna grep for in the logs.
   * e.g., Maybe it will be useful to grep for the framework name, so maybe we should be shoving that into mesos IDs everywhere (Executor IDs (logviewer & supervisor), Task IDs (logviewer and worker)).
      * *However*, I'm not sure if it's ok to have spaces in these Mesos IDs.  And the Framework Name very well could have spaces.

storm/src/main/storm/mesos/NimbusMesosScheduler.java
changreytang: Correct, at this point the task does not exist anymore, will update the log message.

storm/src/main/storm/mesos/util/MesosCommon.java
changreytang: @erikdw I took out the spaces because if splitting on '|' it would cause unneeded spaces on both sides

storm/src/main/storm/mesos/util/MesosCommon.java
changreytang: @erikdw If the mesos component name delimiter is configurable then we should probably make this use `MesosCommon.getMesosComponentNameDelimiter()`. My question is, what is the advantage of making it configurable in the first place? This gives the user ability to possibly break storm-mesos.

storm/src/main/storm/mesos/util/MesosCommon.java
erikdw: That's easily solvable by trimming the resultant strings.  But I'm not sure if spaces in mesos IDs is a bad idea or not

storm/src/main/storm/mesos/util/MesosCommon.java
erikdw: Name and ID are different things.  The configurability only relates to the name, which is a human readable nice string.  We shouldn't have any logic dependent on the format of the name itself.

storm/src/main/storm/mesos/util/MesosCommon.java
changreytang: @erikdw Essentially, `MESOS_COMPONENT_ID_DELIMITER = '|'` and leave `DEFAULT_MESOS_COMPONENT_NAME_DELIMITER = ' | '`

storm/src/main/storm/mesos/MesosNimbus.java
changreytang: @erikdw So you would want this framework to create this directory in ZK if it doesn't exist at startup? I see how this could be useful in the future if we need a place to store state.

storm/src/main/storm/mesos/util/MesosCommon.java
erikdw: Yeah, that's what I'm thinking.

storm/src/main/storm/mesos/MesosNimbus.java
erikdw: I haven't fully read every line of code yet (my initial review was focusing on touchpoints that I was already familiar with), but I'm assuming that anything needing to create a file in ZK would go through a `mkdir -p` type of operation in ZK to ensure the parent dir already exists.  So I wouldn't expect an explicit and separate "create the root dir in ZK" piece of logic.  I would expect that to happen naturally whenever *anything* goes to store state in ZK for storm-mesos itself.

changreytang: I actually don't use that type of operation for my implementation of creating nodes in ZK because I didn't know they had it. Thanks for the heads up, I'll add it now.

erikdw: Though I should state that I'm not a ZK API-usage expert by any means, so it might not be natural or easy to do it this way.

storm/src/main/storm/mesos/NimbusMesosScheduler.java
```diff
@@ -109,6 +119,28 @@ public void statusUpdate(SchedulerDriver driver, TaskStatus status) {
...
 
+  private void updateLogviewerState(TaskStatus status) {
+    String nodeId = status.getTaskId().getValue().split("\\" + DEFAULT_MESOS_COMPONENT_NAME_DELIMITER)[0];
+    switch (status.getState()) {
+      case TASK_STAGING: return;
+      case TASK_STARTING: return;
+      case TASK_RUNNING: return;
+      default:
+    }
+    // if it gets to this point it means logviewer terminated; update ZK with new logviewer state
```
changreytang: I see, in the case of task status 10, 12, and 13 the task may or may not still be running due to some type of network partitioning or what not. Luckily the driver has an explicit kill command we can use.

storm/src/main/storm/mesos/MesosNimbus.java
```diff
@@ -335,6 +378,117 @@ public boolean isHostAccepted(String hostname) {
...
+                              .setTaskId(taskId)
+                              .setName(name)
+                              .setSlaveId(aggregatedOffers.getSlaveID())
+                              .setCommand(commandInfoBuilder.build())
+                              .addAllResources(resources)
+                              .build();
+
+      logviewerTask.add(task);
+
+      LOG.info("launchLogviewer: Using offerIDs: {} on host: {} to launch logviewer", offerIDListToString(offerIDList), nodeId);
```
erikdw: add taskId into the log line?

storm/src/main/storm/mesos/NimbusMesosScheduler.java
```diff
@@ -109,6 +119,48 @@ public void statusUpdate(SchedulerDriver driver, TaskStatus status) {
     }
   }
 
+  private void updateLogviewerState(TaskStatus status) {
+    String taskId = status.getTaskId().getValue();
+    if (!taskId.contains(MesosCommon.MESOS_COMPONENT_ID_DELIMITER)) {
+      LOG.error("updateLogviewerState: taskId for logviewer, {}, isn't formatted correctly", taskId);
```
erikdw: maybe add "so ignoring task update"

storm/src/main/storm/mesos/NimbusMesosScheduler.java
```diff
@@ -109,6 +119,48 @@ public void statusUpdate(SchedulerDriver driver, TaskStatus status) {
...
+        return;
+      case TASK_RUNNING:
+        checkRunningLogviewerState(logviewerZKPath);
+        return;
+      case TASK_LOST:
+        // this status update can be triggered by the explicit kill and isn't terminal, do not kill again
+        break;
+      default:
+        // explicitly kill the logviewer task to ensure logviewer is terminated
+        mesosNimbus._driver.killTask(status.getTaskId());
```
erikdw: Now I'm a bit scared of this behavior. :)  Hopefully we won't have a combination of logviewers failing to be able to launch along with a cyclical task update behavior like we saw (i.e., receive TASK_LOST -> call killTask -> receive another TASK_LOST -> ...).

@changreytang @JessicaLHartog what is your feeling about this?

storm/src/main/storm/mesos/util/ZKClient.java
```diff
@@ -0,0 +1,117 @@
...
+    } catch (Exception e) {
+      LOG.error(e.toString());
+      return false;
+    }
+  }
+
+  public boolean createNode(String path) {
+    try {
+      // default payload of byte[0]
+      _client.create().creatingParentsIfNeeded().forPath(path);
```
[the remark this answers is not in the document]
changreytang: @erikdw A ZK Node is a combination of a path name and some type of data payload (in bytes). I was just specifying that for this createNode() method, you will not need to specify any data to stuff into the ZK Node and it will automatically default to byte[0] which is pretty much nothing

storm/src/main/storm/mesos/util/ZKClient.java
```diff
@@ -0,0 +1,117 @@
...
+                                     .sessionTimeoutMs(sessionTimeout)
+                                     .build();
+    _client.start();
+  }
+
+  public boolean createNode(String path, String data) {
+    try {
+      _client.create().creatingParentsIfNeeded().forPath(path, data.getBytes());
+      return true;
+    } catch (Exception e) {
```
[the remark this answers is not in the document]
changreytang: @erikdw It seems like the forPath() method throws a generic Exception so I need to catch that. For reference:
```java
package org.apache.curator.framework.api;

public interface PathAndBytesable<T> {
    T forPath(String var1, byte[] var2) throws Exception;
    T forPath(String var1) throws Exception;
}
```

storm/src/main/storm/mesos/NimbusMesosScheduler.java
changreytang: @erikdw Personally I don't think the cycle would happen again after we handled the TASK_LOST specifically. The use case for this line of code is that we get a non-terminal task death update (caused by a network partition) in which the logviewer is actually still running so we kill it explicitly such that we won't accidentally launch two logviewers on the same host. 

SIDE NOTE: I think a good check to eventually put in before launching a logviewer on a host is to check that the logviewer port is free. Currently, the logviewer process itself doesn't care if the port is already taken and still runs, which is the reason the logviewer process flood is able to happen. If we put a check of the port before launching a logviewer on a host I think that would be helpful check for ensuring that only one logviewer is running on each host.

storm/src/main/storm/mesos/util/ZKClient.java
```diff
@@ -0,0 +1,117 @@
...
+      LOG.error(e.toString());
+      return false;
+    }
+  }
+
+  public void deleteNode(String path) {
+    try {
+      // the delete is guaranteed even if initial failure, curator will attempt to delete in background until successful
+      _client.delete().guaranteed().forPath(path);
```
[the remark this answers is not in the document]
changreytang: @erikdw If there is a logviewer task running that ZK doesn't know about that will be eventually reconciled. But currently, if ZK thinks there is a logviewer running but it is actually dead (aka if the Nimbus dies then the logviewer dies) then we will not reconcile that state because we don't have "heartbeats"

SUGGESTION: If we track the TaskIDs of logviewers somewhere (ZK, List<TaskID>, etc.) then we can reconcile that state (through task reconciliation) by passing it the logviewer TaskIDs from which we would receive TASK_LOST if they were actually dead.

storm/src/main/storm/mesos/MesosNimbus.java
```diff
@@ -203,13 +218,29 @@ void initializeMesosStormConf(Map conf, String localDir) {
...
+    LOG.info("Logviewer information will be stored under {}", _logviewerZkDir);
+
+    if (zooKeeperPort == null || zooKeeperServers == null) {
+      LOG.error("ZooKeeper configs are not found in storm.yaml");
+    } else {
+      StringBuilder connectionString = new StringBuilder();
+      for (String server : zooKeeperServers) {
+        connectionString.append(String.format("%s:%s,", server, zooKeeperPort));
+      }
+      _zkClient = new ZKClient(connectionString.substring(0, connectionString.length() - 1));
```
JessicaLHartog: This might be better as: `_zkClient = new ZKClient(connectionString.toString());`

storm/src/main/storm/mesos/MesosNimbus.java
```diff
@@ -249,6 +280,32 @@ public void doRegistration(final SchedulerDriver driver, Protos.FrameworkID id)
...
+        // in the framework scheduler's statusUpdate() method
+        Collection<TaskStatus> taskStatuses = new ArrayList<TaskStatus>();
+        List<String> logviewerPaths = _zkClient.getChildren(_logviewerZkDir);
+        if (logviewerPaths == null) {
+          _driver.reconcileTasks(taskStatuses);
+          return;
+        }
+        for (String path : logviewerPaths) {
+          TaskID logviewerTaskId = TaskID.newBuilder()
+                                                .setValue(new String(_zkClient.getNodeData(String.format("%s/%s", _logviewerZkDir, path))))
```
JessicaLHartog: I think the alignment got messed up here, as well as in the `TaskStatus` creation below this.

storm/src/main/storm/mesos/MesosNimbus.java
changreytang: @JessicaLHartog I used substring( ) because I was scared the extra appended ',' at the end of the connection string would screw things up

storm/src/main/storm/mesos/util/ZKClient.java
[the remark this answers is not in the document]
erikdw: @changreytang : yup, seems like we have no choice here.

storm/src/main/storm/mesos/util/ZKClient.java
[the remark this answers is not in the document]
erikdw: @changreytang : I'll try to put a clarifying comment in, I still don't really get it.

storm/src/main/storm/mesos/util/ZKClient.java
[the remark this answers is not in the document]
erikdw: @changreytang : thanks for implementing the SUGGESTION, per our offline discussions!

storm/src/main/storm/mesos/NimbusMesosScheduler.java
erikdw: @changreytang : yup, I heard what you are saying about checking the logviewer port.  I'm not sure how to do that nicely, maybe a wrapper / initial cmd to be executed which needs to succeed before we launch the logviewer.  Where that cmd could be something *like*: 

```
% sudo netstat -anpl | grep -e ':8080\s\s*.*LISTEN'
tcp        0      0 :::8080                     :::*                        LISTEN      22916/java  
```

storm/src/main/storm/mesos/MesosNimbus.java
erikdw: We've changed this logic around a bit in our latest "minor cleanup commit".

storm/src/main/storm/mesos/NimbusMesosScheduler.java
erikdw: Issue #218 filed

storm/src/main/storm/mesos/util/ZKClient.java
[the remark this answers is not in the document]
erikdw: My clarifying comment: https://github.com/mesos/storm/commit/84c34faae55dfbebb2343f0d83d3a40340c08f7b

storm/src/main/storm/mesos/MesosNimbus.java
erikdw: Here's the commit where we tweaked the logic:
* https://github.com/mesos/storm/commit/72681481f4df0ab33da08f7b6e9d9ba3db17c85e

PR #223, opened 2017-10-19
Update task name to better support multiple Storm frameworks running …

JessicaLHartog:
…side-by-side on Mesos


JessicaLHartog:
Actually, this will be addressed in #209 so closing this as it is invalid.

PR #237, opened 2018-03-13
upgrade storm from 1.0.5 to 1.0.6

erikdw:
* Also remove temporary use of reflection to invoke Supervisor.launchDaemon() since that is no longer needed with storm 1.0.6
* Also change default mesos versions in the pom.xml to 0.28.2 instead of 0.27.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,8 @@ script:
 matrix:
   fast_finish: true
   include:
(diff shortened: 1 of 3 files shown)


erikdw:
@JessicaLHartog : please review when you have a chance!

erikdw:
@JessicaLHartog : thx!

PR #240, opened 2018-03-17
Reap previously launched logviewer tasks

### LINKED ISSUE #239: Nimbus should kill already-launched logviewers if relaunched with `sidecar.enabled=false`
Nimbus should kill already-launched logviewers if re-launched with `mesos.logviewer.sidecar.enabled=false`.

The pertinent use case is: If Nimbus is initially launched with `mesos.logviewer.sidecar.enabled=true` (or absent, which will give it the default `true` value) and then re-launched with `mesos.logviewer.sidecar.enabled=false`. logviewers launched prior to the re-launch are never reaped. Nimbus should reap them and release the resources back to Mesos.

srikanth-viswanathan:
Reap previously launched logviewer tasks if `mesos.logviewer.sidecar.enabled` is now `false`

Fixes #239

### DIFF
--- a/storm/src/main/storm/mesos/MesosNimbus.java
+++ b/storm/src/main/storm/mesos/MesosNimbus.java
@@ -43,11 +43,11 @@ import org.apache.mesos.Protos.OfferID;
 import org.apache.mesos.Protos.Resource;
 import org.apache.mesos.Protos.TaskID;
 import org.apache.mesos.Protos.TaskInfo;
+import org.apache.mesos.Protos.TaskState;
 import org.apache.mesos.Protos.TaskStatus;
 import org.apache.mesos.Protos.Value.Range;
 import org.apache.mesos.Protos.Value.Ranges;
 import org.apache.mesos.Protos.Value.Scalar;
-import org.apache.mesos.Protos.TaskState;
 import org.apache.mesos.SchedulerDriver;
 import org.json.simple.JSONValue;
 import org.slf4j.Logger;
@@ -60,8 +60,8 @@ import storm.mesos.resources.ResourceEntries.ScalarResourceEntry;
 import storm.mesos.resources.ResourceEntry;
 import storm.mesos.resources.ResourceNotAvailableException;
 import storm.mesos.resources.ResourceType;
-import storm.mesos.schedulers.StormSchedulerImpl;
 import storm.mesos.schedulers.IMesosStormScheduler;
+import storm.mesos.schedulers.StormSchedulerImpl;
 import storm.mesos.shims.CommandLineShimFactory;
 import storm.mesos.shims.ICommandLineShim;
 import storm.mesos.shims.LocalStateShim;
@@ -91,8 +91,8 @@ import java.util.Timer;
 import java.util.TimerTask;
 
 import static storm.mesos.util.PrettyProtobuf.offerIDListToString;
-import static storm.mesos.util.PrettyProtobuf.offerToString;
 import static storm.mesos.util.PrettyProtobuf.offerMapToString;
+import static storm.mesos.util.PrettyProtobuf.offerToString;
 import static storm.mesos.util.PrettyProtobuf.taskInfoListToString;
 import static storm.mesos.util.PrettyProtobuf.taskStatusListToTaskIDsString;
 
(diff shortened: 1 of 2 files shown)


srikanth-viswanathan:
The only side effect of this change is that you'll now see this line every 5 minutes:

```
2018-03-17 03:25:16.968 s.m.u.ZKClient [ERROR] org.apache.zookeeper.KeeperException$NoNodeException: KeeperErrorCode = NoNode for /storm-mesos/logviewers
```
Do you think this is a problem? I'm open to your suggestions.

storm/src/main/storm/mesos/NimbusMesosScheduler.java
```diff
@@ -17,13 +17,7 @@
...
 package storm.mesos;
 
-import org.apache.mesos.Protos.ExecutorID;
-import org.apache.mesos.Protos.FrameworkID;
-import org.apache.mesos.Protos.MasterInfo;
-import org.apache.mesos.Protos.Offer;
-import org.apache.mesos.Protos.OfferID;
-import org.apache.mesos.Protos.SlaveID;
-import org.apache.mesos.Protos.TaskStatus;
+import org.apache.mesos.Protos.*;
```
JessicaLHartog: Can you revert the * import. We don't like those.

storm/src/main/storm/mesos/NimbusMesosScheduler.java
```diff
@@ -98,6 +94,7 @@ public void statusUpdate(SchedulerDriver driver, TaskStatus status) {
     if (status.getTaskId().getValue().contains("logviewer")) {
       updateLogviewerState(status);
     }
+
```
JessicaLHartog: Please delete newly introduced newlines here at line 97 and at 125, 128, and 134 as well.

storm/src/main/storm/mesos/NimbusMesosScheduler.java
```diff
@@ -154,13 +160,34 @@ private void updateLogviewerState(TaskStatus status) {
     }
   }
 
-  private void checkRunningLogviewerState(String logviewerZKPath) {
+  private void ensureZNodeExists(String logviewerZKPath) {
```
JessicaLHartog: Since this is more for the Logviewer ZNode, can we rename it from `ensureZNodeExists` to `ensureLogviewerZnodeExists`?

storm/src/main/storm/mesos/MesosNimbus.java
```diff
@@ -286,34 +288,32 @@ public void doRegistration(final SchedulerDriver driver, Protos.FrameworkID id)
...
-            TaskID logviewerTaskId = TaskID.newBuilder()
-                                           .setValue(new String(_zkClient.getNodeData(String.format("%s/%s", _logviewerZkDir, path))))
-                                           .build();
-            TaskStatus logviewerTaskStatus = TaskStatus.newBuilder()
-                                                       .setTaskId(logviewerTaskId)
-                                                       .setState(TaskState.TASK_RUNNING)
-                                                       .build();
-            taskStatuses.add(logviewerTaskStatus);
-          }
+    _timer.scheduleAtFixedRate(new TimerTask() {
```
JessicaLHartog: align this with the rest of the code please

storm/src/main/storm/mesos/MesosNimbus.java
```diff
@@ -286,34 +288,32 @@ public void doRegistration(final SchedulerDriver driver, Protos.FrameworkID id)
...
+      @Override
+      public void run() {
+        // performing "explicit" reconciliation; master will respond with the latest state for all logviewer tasks
+        // in the framework scheduler's statusUpdate() method
+        List<TaskStatus> taskStatuses = new ArrayList<TaskStatus>();
+        List<String> logviewerPaths = _zkClient.getChildren(_logviewerZkDir);
```
JessicaLHartog: Is this the line of code that is generating the log line that you mention as the visible side-effect of this change?

storm/src/main/storm/mesos/MesosNimbus.java
srikanth-viswanathan: This is actually aligned with the rest of the code. The diff makes it seem like it isn't. I'm taking the timer thread out of the `if (_enabledLogviewerSidecar)` block and scheduling it unconditionally. See https://github.com/srikanth-viswanathan/mesos-storm/blob/66dcbbe502aa3438b2a1b107d146e25de058f82c/storm/src/main/storm/mesos/MesosNimbus.java#L291 for the full file.

storm/src/main/storm/mesos/MesosNimbus.java
srikanth-viswanathan: Correct. The more I think about this (and having seen the code run on our systems), I think this trace will be a nuisance, given how frequently the timer task runs (every 5 mins). I'm in favor of adding an explicit check to first test that the `logviewer` ZNode exists (without the error trace) before trying to get its children. Thoughts?

storm/src/main/storm/mesos/NimbusMesosScheduler.java
srikanth-viswanathan: Sure, I didn't realize my IntelliJ settings had that enabled by default.

srikanth-viswanathan:
@JessicaLHartog Thanks for the review. I've pushed changes that address your comments. 

P.S.: I had originally added spaces around lines of code to add visual separation for readability. I've since removed them after your comments, but I'm curious about the reason for your suggestion.

storm/src/main/storm/mesos/MesosNimbus.java
JessicaLHartog: Sorry for the incredibly slow response. I like this solution, please implement.

PR #243, opened 2018-07-23
using new storm release - avoid kidnapped workers

abhinavmehta14:
[Release](https://github.com/erikdw/storm/releases/tag/v1.0.6-storm-mesos3) for Storm version used here.


erikdw:
:shipit:  Thanks!

